PR #33, opened 2020-05-19
Solution

Andreas-Just:


src/api/getMovies.ts
```diff
@@ -0,0 +1,9 @@
+const API_KEY_OMDB = process.env.REACT_APP_API_KEY_OMDB;
```
irenhh, line 1: 👍

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,55 +1,159 @@
...
+    setError('');
+  };
+
+  const handleFormSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    findMovie();
+  };
+
+  const handleClickSubmit = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
```
irenhh, line 95: doesn't it work without preventDefault method?

PR #76, opened 2020-09-16
done

DmJSCourse:
https://dmjscourse.github.io/react_movies-list-fetch-movies/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,23 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (movie) => {
+    if (this.state.movies.find(element => element.imdbId === movie.imdbID)) {
+      return;
+    }
+
+    this.setState(prevState => ({
+      movies: [...prevState.movies,
+        {
+          title: movie.Title,
+          description: movie.Plot,
+          imgUrl: movie.Poster,
+          imdbUrl: `https://www.imdb.com/title/${movie.imdbID}`,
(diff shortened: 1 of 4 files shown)


src/api/api.js
```diff
@@ -0,0 +1,7 @@
+const URL = 'http://www.omdbapi.com/?apikey=e6b7e90f';
```
rmnkk, line 1: You should change the URL to https. Because GitHub doesn't allow requests to http.

src/App.js
```diff
@@ -9,6 +9,23 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (movie) => {
+    if (this.state.movies.find(element => element.imdbId === movie.imdbID)) {
+      return;
+    }
+
+    this.setState(prevState => ({
```
rmnkk, line 17: Yous should use hooks instead of class components

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,91 @@
...
-            className="button is-light"
-          >
-            Find a movie
-          </button>
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
```
rmnkk, line 26: This class should only appear when there is an error.

rmnkk (changes requested):

PR #136, opened 2021-01-28
solution

Pasha5555:
[DEMO LINK](https://pasha5555.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,33 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
-import data from './api/movies.json';
 
-export class App extends Component {
-  state = {
-    movies: data,
+export const App = () => {
+  const [movies, setMovie] = useState([]);
+
+  const addMovie = (findedMovie) => {
+    setMovie((availableMovies) => {
+      const moviesForCompare = availableMovies
+        .filter(movie => movie.imdbID !== findedMovie.imdbID);
+
+      if (availableMovies.length === moviesForCompare.length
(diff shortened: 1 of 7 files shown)


zhenya-mezhueva (approved):

src/App.js
```diff
@@ -1,26 +1,33 @@
...
+    setMovie((availableMovies) => {
+      const moviesForCompare = availableMovies
+        .filter(movie => movie.imdbID !== findedMovie.imdbID);
 
-  render() {
-    const { movies } = this.state;
+      if (availableMovies.length === moviesForCompare.length
+        && findedMovie.imdbID) {
+        return [...availableMovies, findedMovie];
+      }
```
varseniuk, line 17: You can just check if there's SOME movie in the list using the array method 'some'

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,97 @@
...
+  const [inputValue, setValue] = useState('');
+  const [movie, setMovie] = useState('');
+
+  const fetchMovie = async() => {
+    const res = await getMovies(inputValue);
+
+    setMovie(res);
+  };
+
+  const addFindedMovie = (e) => {
```
varseniuk, line 20: Why do you need 'event' here?

varseniuk (changes requested):
The task says:

src/components/FindMovie/FindMovie.js
Pasha5555: Because I use e.preventDefault() to correct work form event onSubmit. Without it page just update without adding movie to the list

varseniuk: 👍

varseniuk (approved):

PR #150, opened 2021-01-30
Develop

Layrovell:
[DEMO LINK](https://Layrovell.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,109 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
-
+import PropTypes from 'prop-types';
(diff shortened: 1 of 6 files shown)


src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,109 @@
...
+                if (!query) {
+                  setIsValidUrl(true);
+
+                  return;
+                }
+
+                props.addMovie(movie);
+                setQuery('');
+                setMoviePreview(false);
+              }}
```
varseniuk, line 87: let's make a separate function for this

varseniuk (changes requested):
The task says:

varseniuk (approved):
Good job  👍

PR #153, opened 2021-01-30
Develop

dtsehelnyk:
[DEMO LINK](https://dtsehelnyk.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,31 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import data from './api/movies.json';
 
-export class App extends Component {
-  state = {
-    movies: data,
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (newMovie) => {
+    if (!newMovie || movies.find(movie => movie.imdbID === newMovie.imdbID)) {
+      return;
+    }
+
+    setMovies([
+      ...movies,
+      newMovie,
+    ]);
   };
 
-  render() {
-    const { movies } = this.state;
-
-    return (
-      <div className="page">
-        <div className="page-content">
-          <MoviesList movies={movies} />
-        </div>
-        <div className="sidebar">
-          <FindMovie />
-        </div>
+  return (
+    <div className="page">
+      <div className="page-content">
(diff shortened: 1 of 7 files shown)


src/App.js
```diff
@@ -1,26 +1,31 @@
...
 
-export class App extends Component {
-  state = {
-    movies: data,
-  };
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (newMovie) => {
+    if (!newMovie || movies.find(movie => movie.imdbID === newMovie.imdbID)) {
```
varseniuk, line 11: Let's use method 'some' for this

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,108 @@
...
+export const FindMovie = ({ addMovie }) => {
+  const key = '85d016e4';
+
+  const [enteredName, setMovieSearch] = useState('');
+  const [foundMovie, setMovie] = useState(null);
+  const [isCorrect, setAvailability] = useState(false);
+
+  const movieRequest = async() => {
+    try {
+      // eslint-disable-next-line
```
varseniuk, line 17: Why?

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,108 @@
...
+      const find = await fetch(`https://www.omdbapi.com/?apikey=${key}&t=${enteredName}`);
```
varseniuk, line 18: You should do this in api.js file

varseniuk (changes requested):
The task says:

Also, initially, when page is loaded, there is ALREADY error message under input.
And after successfully finding the movie and adding it ti the List, error also shows immediately
Fix this

src/App.js
```diff
@@ -1,26 +1,31 @@
...
 
-export class App extends Component {
-  state = {
-    movies: data,
-  };
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (newMovie) => {
+    if (!newMovie || movies.some(movie => movie.imdbID === newMovie.imdbID)) {
```
varseniuk, line 11:
```suggestion
    if (movies.some(movie => movie.imdbID === newMovie.imdbID)) {
```
You always pass movie to this function in your code, so there's no need to check if it is present

varseniuk (approved):
Very good 👍

PR #155, opened 2021-02-01
solution

sashalarin202:
[DEMO LINK](https://sashalarin202.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,18 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (newMovie) => {
+    const { movies } = this.state;
+    const isMovieInData = movies
+      .every(movie => movie.imdbId !== newMovie.imdbId);
+
+    if (isMovieInData && newMovie) {
+      this.setState(prevState => ({
+        movies: [...prevState.movies, newMovie],
+      }));
+    }
+  }
+
   render() {
     const { movies } = this.state;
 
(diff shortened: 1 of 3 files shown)


andrew-zaycev (approved):
Well done!

src/App.js
```diff
@@ -9,6 +9,18 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (newMovie) => {
+    const { movies } = this.state;
```
vsm2k17, line 13: it would be nice to use a functional component and hooks here

src/App.js
```diff
@@ -9,6 +9,18 @@ export class App extends Component {
...
+    const isMovieInData = movies
```
vsm2k17, line 14: variable name a bit confusing

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,103 @@
...
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie = ({ addMovie }) => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState('');
+  const [movieNotFound, setMovieNotFound] = useState(false);
+
+  const findMovie = async() => {
+    if (query) {
```
vsm2k17, line 14: It looks better if use something like this
```
if (!value) {
  return
}
```
in such a way you can avoid nesting if blocks

vsm2k17 (approved):
good job, don't forget to fix comments

PR #159, opened 2021-02-02
Solution

ShyianO:
[DEMO LINK](https://ShyianO.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,29 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
+
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import data from './api/movies.json';
 
-export class App extends Component {
-  state = {
-    movies: data,
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (movie) => {
+    if (movies.find(item => item.imdbID === movie.imdbID)) {
+      return;
+    }
+
+    setMovies([...movies, movie]);
   };
 
-  render() {
-    const { movies } = this.state;
-
-    return (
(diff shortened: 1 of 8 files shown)


src/api/api.js
```diff
@@ -0,0 +1,11 @@
+const BASE_URL = 'http://www.omdbapi.com/?apikey=5ba996a5&t=';
```
varseniuk, line 1:
```suggestion
const BASE_URL = 'https://www.omdbapi.com/?apikey=5ba996a5&t=';
```
Without it your app will not work properly on github pages

src/App.js
```diff
@@ -1,26 +1,29 @@
...
-  state = {
-    movies: data,
-  };
+export const App = () => {
+  const [movies, setMovies] = useState(data);
 
-  render() {
-    const { movies } = this.state;
+  const addMovie = (movie) => {
+    if (movies.find(item => item.imdbID === movie.imdbID)) {
```
varseniuk, line 12: let's use method 'some' for this

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,104 @@
...
-          />
-        </div>
+import { getMovie } from '../../api/api';
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie = ({ onAddMovie }) => {
+  const [value, setValue] = useState('');
```
varseniuk, line 10: Let's name this variable more meaningful. It's hard to understand, what is 'value'

varseniuk (changes requested):
Check your demolink - it doesn't work correctly.
Change http to https in your URL.
Also, pay attention to my comments

PR #161, opened 2021-02-02
Fetch movies	

kerpichonka:
[DEMO LINK](https://kerpichonka.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,29 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import data from './api/movies.json';
 
-export class App extends Component {
-  state = {
-    movies: data,
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (movie) => {
+    const hasSameMovie = movies
+      .some(film => film.imdbId === movie.imdbId);
+
+    if (!hasSameMovie) {
+      setMovies([...movies, movie]);
+    }
   };
 
-  render() {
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -1,26 +1,29 @@
...
-export class App extends Component {
-  state = {
-    movies: data,
-  };
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const addMovie = (movie) => {
+    const hasSameMovie = movies
+      .some(film => film.imdbId === movie.imdbId);
```
varseniuk, line 12: No need to create this duplicate. Just check if (movies.some....) and if there's already such movie in the list - do nothing (return) else - add it to the list.

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,110 @@
...
-        </div>
+import { getMovies } from '../../data/api';
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie = ({ addMovie }) => {
+  const [title, setTitle] = useState('');
+  const [movie, setMovie] = useState('');
```
varseniuk, line 10: 'movie' is an object. You can initially set it to null or {}, but definitely not to empty string

varseniuk (changes requested):
Read the task description with more attention:
If film is not found show an error message below the input. **Hide the error after changing a title**. - now your error still is shown when you're changing the input.

Add a movie to the list after submitting a form and clear the input. -- now your input is not clearing after adding a new movie

Also, your input is always has a red outline (class 'is-danger') though it should only become red if there's an error.

Also pay attention to my comments

PR #183, opened 2021-03-26
Dev

Ruslanwp:
[🕷](https://Ruslanwp.github.io/react_movies-list-fetch-movies/)


src/api/helpers.js
```diff
@@ -0,0 +1,5 @@
+const BASE_URL = 'http://www.omdbapi.com/?apikey=41fc156d&t=';
```
bogdandobak, line 1: image
```suggestion
const BASE_URL = 'https://www.omdbapi.com/?apikey=41fc156d&t=';
```

bogdandobak (changes requested):
Don't work

PR #213, opened 2021-06-02
Solution

Burunduk003:
- [DEMO LINK](https://Burunduk003.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,106 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
 
 import { MovieCard } from '../MovieCard';
-import movies from '../../api/movies.json';
+import { getMovie } from '../../api/api';
 
-export const FindMovie = () => (
-  <>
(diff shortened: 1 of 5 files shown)


src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
+  const movieSearch = (name) => {
+    getMovie(name)
+      .then((data) => {
+        setMovie({
+          title: data.Title,
+          description: data.Plot,
+          imgUrl: data.Poster,
+          imdbUrl: `https://www.imdb.com/title/${data.imdbID}/`,
+          imdbId: data.imdbID,
+        });
```
Ihor-Karpyn, line 23: I think will better take normalize data in your API file, because if server system will change you must solve mistake in a lot of places

Ihor-Karpyn (approved):
Good and clear solution!!!

src/components/FindMovie/FindMovie.js
VinnieJ-2k20: Don't agree here, api usually doesn't know in what format do you need data on frontend. You get what you get and then have to work with it.

VinnieJ-2k20 (approved):
Good job 👍

PR #219, opened 2021-06-03
Solution

TimurHrynenko:
- [DEMO LINK](https://TimurHrynenko.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,16 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (newMovie) => {
+    const { movies } = this.state;
+
+    if (!movies.find(movie => movie.imdbId === newMovie.imdbId)) {
+      this.setState(state => ({
+        movies: state.movies.concat(newMovie),
+      }));
+    }
+  }
+
   render() {
     const { movies } = this.state;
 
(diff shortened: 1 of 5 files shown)


victor-arutiunov (commented):
Nice work!

src/App.js
```diff
@@ -7,18 +7,43 @@ import data from './api/movies.json';
 export class App extends Component {
```
VinnieJ-2k20, line 7: > Don't use class components. Use React Hooks instead.

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,154 @@
...
+
+  const [movie, setMovie] = useState({
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+  });
+
+  const [titleError, setTitleError] = useState({
+    staus: false,
```
VinnieJ-2k20, line 23:
```suggestion
    status: false,
```

VinnieJ-2k20 (changes requested):
You need to rewrite one more class component to functional with hooks.
Also, I would change imdb Id to title here in message, because we search by title.
image

VinnieJ-2k20 (approved):
👍

PR #220, opened 2021-06-03
Develop

avgusta92:
- [DEMO LINK](https://avgusta92.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,18 @@ export class App extends Component {
     movies: data,
   };
 
+  addNewMovie = (movie) => {
+    this.setState((state) => {
+      const hadMovie = state.movies.some(el => el.imdbId === movie.imdbId);
+
+      return {
+        movies: !hadMovie
+          ? [...state.movies, movie]
+          : [...state.movies],
+      };
+    });
+  }
+
   render() {
     const { movies } = this.state;
 
@@ -18,7 +30,7 @@ export class App extends Component {
           <MoviesList movies={movies} />
         </div>
         <div className="sidebar">
-          <FindMovie />
+          <FindMovie addNewMovie={this.addNewMovie} />
         </div>
       </div>
     );
--- a/src/api/getMovies.js
+++ b/src/api/getMovies.js
@@ -1,0 +1,9 @@
+const reqvest = (endPoint) => {
+  const yourkey = '1c07d812';
+  const URL = `https://www.omdbapi.com/?apikey=${yourkey}&t=${endPoint}`;
+
+  return fetch(URL)
+    .then(response => response.json());
+};
+
+export const getMovie = title => reqvest(title);
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,111 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import PropTypes from 'prop-types';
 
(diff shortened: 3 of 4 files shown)


src/App.js
```diff
@@ -9,6 +9,18 @@ export class App extends Component {
     movies: data,
   };
 
+  addNewMovie = (movie) => {
+    this.setState((state) => {
+      const hadMovie = state.movies.some(el => el.imdbId === movie.imdbId);
+
```
iryna-vys, line 15: I think it's better to do this check before setState, and if movies include movie don't call setState.

src/api/getMovies.js
```diff
@@ -0,0 +1,9 @@
+const reqvest = (endPoint) => {
```
iryna-vys, line 1: Missprint request

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,111 @@
...
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
+import { getMovie } from '../../api/getMovies';
+
+export const FindMovie = ({ addNewMovie }) => {
+  const [qveryValue, setQveryValue] = useState('');
```
iryna-vys, line 9: Missprint query

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,111 @@
...
+              onChange={(event) => {
+                setQveryValue(event.target.value);
+                setErrorMassage('');
+              }}
+            />
+          </div>
+
+          <p className="help is-danger">
+            Can&apos;t find a movie with such a title
+          </p>
```
iryna-vys, line 65: This should be shown by condition. And the className 'is-danger' also.

iryna-vys (approved):
Good!

src/App.js
varseniuk: Yep. Iryna is right

varseniuk: Also, task description says:

src/api/getMovies.js
varseniuk: ➕

src/api/getMovies.js
```diff
@@ -0,0 +1,9 @@
...
+  const yourkey = '1c07d812';
```
varseniuk, line 2:
```suggestion
  const key = '1c07d812';
```

src/components/FindMovie/FindMovie.js
varseniuk: ➕

src/components/FindMovie/FindMovie.js
varseniuk: And again Iryna is right
Снимок экрана 2021-06-07 в 17 11 03

This message is shown by default in your app. It should be vice versa

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,111 @@
...
+              Find a movie
+            </button>
+          </div>
+
+          <div className="control">
+            <button
+              type="button"
+              className="button is-primary"
+              onClick={() => {
+                if (movie !== null) {
```
varseniuk, line 84:
```suggestion
                if (movie) {
```
Actually, there's no need in this check at all

varseniuk (changes requested):
Good start! Let's make some changes and it's done
Left comments

PR #223, opened 2021-06-03
solution

RuslanaPodshyvalova:
[DEMO LINK](https://RuslanaPodshyvalova.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -1,0 +1,24 @@
+const BASE_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=6811da81&';
+
+// eslint-disable-next-line
+const getData = (param) => {
+  return fetch(`${BASE_URL}${param}`)
+    .then((response) => {
+      if (response.ok) {
+        return response.json();
+      }
+
+      throw new Error(`${response.status}: ${response.statusText}`);
+    })
+    .then((response) => {
+      const isError = response?.Error;
+
+      if (!isError) {
+        return response;
+      }
+
+      return null;
+    });
+};
+
+export const getMovie = title => getData(`t=${title}`);
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,155 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
-
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import { getMovie } from '../../api/api';
 import { MovieCard } from '../MovieCard';
-import movies from '../../api/movies.json';
 
-export const FindMovie = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
-        <label className="label" htmlFor="movie-title">
-          Movie title
-        </label>
+export const FindMovie = ({ movies, onAdd }) => {
+  const moviePattern = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+  const [title, setTitle] = useState('');
+  const [movie, setMovie] = useState(moviePattern);
+  const [isMovieCard, setIsMovieCard] = useState(false);
+  const [error, setError] = useState('');
 
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
-        </div>
+  const searchMovie = async() => {
+    if (!title) {
+      return;
+    }
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+    const movieFromServer = await getMovie(title);
 
-      <div className="field is-grouped">
-        <div className="control">
-          <button
-            type="button"
-            className="button is-light"
+    if (movieFromServer !== null) {
+      setIsMovieCard(true);
+      setError('');
+
+      // eslint-disable-next-line
+      const { Title: movieTitle, Plot: description, Poster: imgUrl, imdbID: imdbId } = movieFromServer;
+
+      setMovie({
+        title: movieTitle,
+        description,
+        imgUrl,
+        imdbUrl: `https://www.imdb.com/title/${imdbId}`,
+        imdbId,
+      });
+    } else {
+      setError(`Can&apos;t find a movie with such a title`);
+    }
+  };
+
+  const setUpMovieTitle = (event) => {
+    if (title) {
+      clearState();
+    }
+
+    setTitle(event.target.value);
+  };
+
+  const addMovieToList = () => {
+    clearState();
+    if (isMovieInList()) {
+      setError(`can&apos;t add, this movie is on the list`);
+
+      return;
+    }
+
+    if (movie.title) {
+      onAdd(movie);
+    }
+  };
+
+  const isMovieInList = () => movies.some(curMovie => (
+    curMovie.imdbId === movie.imdbId
+  ));
+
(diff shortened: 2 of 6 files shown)


src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
-        </p>
-      </div>
+export const FindMovie = ({ movies, onAdd }) => {
+  const moviePattern = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
```
YuriiLiso, line 15:
```suggestion
```

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+  const [title, setTitle] = useState('');
+  const [movie, setMovie] = useState(moviePattern);
```
YuriiLiso, line 17:
```suggestion
  const [movie, setMovie] = useState(null);
```
You just can use ```null``` for objects

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+    }
+  };
+
+  const isMovieInList = () => movies.some(curMovie => (
+    curMovie.imdbId === movie.imdbId
+  ));
+
+  const clearState = () => {
+    setTitle('');
+    setMovie(moviePattern);
```
YuriiLiso, line 74:
```suggestion
    setMovie(null);
```

YuriiLiso (approved):
Good looks!

src/api/api.js
```diff
@@ -0,0 +1,24 @@
+const BASE_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=6811da81&';
+
+// eslint-disable-next-line
```
varseniuk, line 3: let's not disable es-lint

src/api/api.js
```diff
@@ -0,0 +1,24 @@
...
+    .then((response) => {
+      const isError = response?.Error;
+
+      if (!isError) {
+        return response;
+      }
+
+      return null;
+    });
+};
```
varseniuk, line 22:
```suggestion
const getData = (param) => {
  return fetch(`${BASE_URL}${param}`)
    .then((response) => {
      if (!response.ok) {
        throw new Error(`${response.status}: ${response.statusText}`);
      }
      
      return responce.json();  
    })
```
Should be enough

src/components/FindMovie/FindMovie.js
varseniuk: Yurii is right. You don't need this. It easily can be `null`

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+  const moviePattern = {
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  };
+  const [title, setTitle] = useState('');
+  const [movie, setMovie] = useState(moviePattern);
+  const [isMovieCard, setIsMovieCard] = useState(false);
```
varseniuk, line 18:
```suggestion
```
Actually, you don't need this variable. You can use `movie` variable for conditional rendering. (it may be either null or not null)

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+    }
+
+    const movieFromServer = await getMovie(title);
+
+    if (movieFromServer !== null) {
+      setIsMovieCard(true);
+      setError('');
+
+      // eslint-disable-next-line
+      const { Title: movieTitle, Plot: description, Poster: imgUrl, imdbID: imdbId } = movieFromServer;
```
varseniuk, line 33:
```suggestion
      const {
        Title: movieTitle,
        Plot: description,
        Poster: imgUrl,
        imdbID: imdbId
      } = movieFromServer;
```

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+              className="input is-danger"
+              value={title}
+              onChange={event => setUpMovieTitle(event)}
+            />
+          </div>
+
+          <p className={classnames({
+            help: true,
+            's-danger': true,
+            hide: error.length === 0,
```
varseniuk, line 101:
```suggestion
            hide: !error,
```
But let's better use conditional rendering

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+            </button>
+          </div>
+
+          <div className="control">
+            <button
+              type="button"
+              className={classnames({
+                button: true,
+                'is-primary': true,
+                hide: !isMovieCard,
```
varseniuk, line 126: use conditional rendering instead

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+            >
+              Add to the list
+            </button>
+          </div>
         </div>
+      </form>
+
+      <div className={classnames({
+        container: true,
+        hide: error.length > 0 || !isMovieCard,
```
varseniuk, line 138:
```suggestion
        hide: !!error || !isMovieCard,
```
Use conditional rendering

src/api/api.js
```diff
@@ -0,0 +1,24 @@
+const BASE_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=6811da81&';
```
varseniuk, line 1:
```suggestion
const BASE_URL = 'https://www.omdbapi.com/?i=tt3896198&apikey=6811da81&';
```
Have you checked your app on gh-pages? You need to change the protocol to `https` to make it work, because now it's not working

https://user-images.githubusercontent.com/66570142/121379554-ce411000-c94c-11eb-8150-a95b1271b92e.mov

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,155 @@
...
+  const [error, setError] = useState('');
+
+  const searchMovie = async() => {
+    if (!title) {
+      return;
+    }
+
+    const movieFromServer = await getMovie(title);
+
+    if (movieFromServer !== null) {
```
varseniuk, line 28:
```suggestion
    if (movieFromServer.Response !== 'False') {
```
As a variant

varseniuk (changes requested):

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,138 @@
...
+      clearState();
+    }
+
+    setTitle(event.target.value);
+  };
+
+  const addMovieToList = () => {
+    clearState();
+    if (isMovieInList()) {
+      setError(`can&apos;t add, this movie is on the list`);
```
varseniuk, line 52: Снимок экрана 2021-06-11 в 09 03 08

Looks not good. Let's fix this

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,138 @@
...
+          <label className="label" htmlFor="movie-title">
+            Movie title
+          </label>
+
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
```
varseniuk, line 85: class `is-danger` should only be set if there's an error. Now your input have red outline by default

varseniuk (approved):
👍 
Pay attention to my comments and try to fix error messages and input's outline

PR #224, opened 2021-06-03
Solution-001

senkiv-oleh:
[DEMO LINK](https://senkiv-oleh.github.io/react_movies-list-fetch-movies/)


Grygoriy-Shytikov (approved):
LGTM! 👍

src/App.js
```diff
@@ -9,6 +9,20 @@ export class App extends Component {
     movies: data,
```
varseniuk, line 9: Let's use functional components and hooks instead of classes (as task requires):

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
-          />
-        </div>
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie = ({ addMovie }) => {
+  const [value, setSearchTerm] = useState('');
+  const [results, setSearchResults] = useState({
```
varseniuk, line 11:
```suggestion
  const [value, setValue] = useState('');
  const [results, setResults] = useState({
```
The name of the function should start with `set` and include the value's name. Actually, you are 'setting' the value

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
+    title: '',
+    description: '',
+    imgUrl: '',
+    imdbUrl: '',
+    imdbId: '',
+  });
+  const [filmNotFound, setNotFound] = useState(false);
```
varseniuk, line 18:
```suggestion
  const [filmNotFound, setFilmNotFound] = useState(false);
```

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
+
+  const handleChange = (event) => {
+    event.preventDefault();
```
varseniuk, line 21:
```suggestion
```
`preventDefault` is only needed to prevent the common behavior on form submitting. For input changes it is redundant and does nothing

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
 
-      <div className="field is-grouped">
-        <div className="control">
-          <button
-            type="button"
-            className="button is-light"
-          >
-            Find a movie
-          </button>
+    if (value === '') {
```
varseniuk, line 40:
```suggestion
    if (!value) {
```

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
+  const [filmNotFound, setNotFound] = useState(false);
+
+  const handleChange = (event) => {
+    event.preventDefault();
+    setSearchTerm(event.target.value);
+    setNotFound(true);
+  };
+
+  const findMovie = () => {
+    const API_URL = `https://www.omdbapi.com/?apikey=1e23943c&t=`;
```
varseniuk, line 27: let's move all api requests to the separate file `api.js`

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,106 @@
...
+          <label className="label" htmlFor="movie-title">
+            Movie title
+          </label>
+
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className={classNames(`input`, { 'is-danger': !filmNotFound })}
```
varseniuk, line 60:
```suggestion
              className={classNames(`input`, { 'is-danger': filmNotFound })}
```
Снимок экрана 2021-06-08 в 10 15 01

Now you have an error message by default even on the app start and it does not work as suppose. Let's fix this

varseniuk (changes requested):
Let's make some small changes and figure out why does your error message work not as supposed now

src/App.js
```diff
@@ -9,6 +9,20 @@ export class App extends Component {
...
   };
 
+  addMovie = (movie) => {
+    const { movies } = this.state;
+    const sameFilm = movies.find(item => item.imdbId === movie.imdbId);
```
aliceblunt8, line 14: or you can use some method, which returns a boolean value

src/api/api.js
```diff
@@ -0,0 +1,6 @@
+export function getMovie(title) {
+  const API_URL = `https://www.omdbapi.com/?apikey=1e23943c&t=`;
```
aliceblunt8, line 2: move this constant out of the function

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,112 @@
...
-        <div className="control">
-          <button
-            type="button"
-            className="button is-light"
-          >
-            Find a movie
-          </button>
+        return Promise.resolve(setFilmFound(false));
+      })
+      .catch(() => setFilmFound(true));
```
aliceblunt8, line 50: are you sure we need to set this value to true?

aliceblunt8 (approved):

PR #225, opened 2021-06-04
solution

vitalya2626:
[DEMO LINK](https://vitalya2626.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,26 +1,26 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import data from './api/movies.json';
 
-export class App extends Component {
-  state = {
-    movies: data,
+export const App = () => {
+  const [movies, setMovies] = useState(data);
+
+  const onAddMovie = (newMovie) => {
+    if (!movies.find(movie => (movie.imdbId === newMovie.imdbId))) {
+      setMovies([...movies, newMovie]);
+    }
   };
 
-  render() {
-    const { movies } = this.state;
-
-    return (
-      <div className="page">
-        <div className="page-content">
-          <MoviesList movies={movies} />
-        </div>
-        <div className="sidebar">
-          <FindMovie />
-        </div>
+  return (
+    <div className="page">
+      <div className="page-content">
+        <MoviesList movies={movies} />
       </div>
-    );
-  }
-}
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -1,26 +1,26 @@
...
-  state = {
-    movies: data,
-  };
+export const App = () => {
+  const [movies, setMovies] = useState(data);
 
-  render() {
-    const { movies } = this.state;
+  const onAddMovie = (newMovie) => {
+    if (!movies.find(movie => (movie.imdbId === newMovie.imdbId))) {
```
varseniuk, line 11: You could use method `some` instead

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,107 @@
...
-      </div>
+export const FindMovie = ({ onAddMovie }) => {
+  const [title, setTitle] = useState('');
+
+  const [movie, setMovie] = useState(null);
+
+  const [movieFound, setMovieFind] = useState(true);
+
+  useEffect(() => {
+    setMovieFind((true));
```
varseniuk, line 16: You could write it on input change

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,107 @@
...
+                input: true,
+                'is-danger': !movieFound,
+              })}
+              value={title}
+              onChange={(event) => {
+                setTitle(event.target.value);
+              }}
+            />
+          </div>
+          {movieFound || (
```
varseniuk, line 58:
```suggestion
          {!movieFound &&(
```
Use `&&` for conditional rendering

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,107 @@
...
+            </button>
+          </div>
+
+          <div className="control">
+            <button
+              type="button"
+              className="button is-primary"
+              onClick={() => {
+                 // eslint-disable-line 
+                movie && onAddMovie(movie);  // eslint-disable-line
```
varseniuk, line 82:
```suggestion
                if (movie) {
                  onAddMovie(movie);
                }
```
let's not disable linter without good reasons for it

varseniuk: Also it would be better to create function and call it here

varseniuk (changes requested):
Good job!
Let's fix an issue with disabling es-lint and it's done

PR #228, opened 2021-06-04
Solution

iryna-vys:
[DEMO LINK](https://iryna-vys.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -9,6 +9,16 @@ export class App extends Component {
     movies: data,
   };
 
+  addMovie = (movie) => {
+    if (this.state.movies.some(({ imdbId }) => imdbId === movie.imdbId)) {
+      return;
+    }
+
+    this.setState(state => ({
+      movies: [...state.movies, movie],
+    }));
+  }
+
   render() {
     const { movies } = this.state;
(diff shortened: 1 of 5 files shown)


avgusta92 (approved):

src/App.js
```diff
@@ -9,6 +9,16 @@ export class App extends Component {
     movies: data,
```
varseniuk, line 9: let's use functional component and hooks instead
Task description says:

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,114 @@
...
-          >
-            Add to the list
-          </button>
+        <div className="field is-grouped">
+          <div className="control">
+            <button
+              type="button"
+              className="button is-light"
+              onClick={() => {
+                getMovie(query)
```
varseniuk, line 60: Let's create a function for this and call it here on click

varseniuk (changes requested):
Good job!
Some small fixes and it's done

varseniuk (approved):
Great job and clean code! Nice

PR #231, opened 2021-06-04
Solution

olha-tsarova:
- [DEMO LINK](https://olha-tsarova.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,103 @@
-import React from 'react';
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
 import './FindMovie.scss';
 
 import { MovieCard } from '../MovieCard';
-import movies from '../../api/movies.json';
+import { findMovie } from '../../api/movies';
 
-export const FindMovie = () => (
-  <>
-    <form className="find-movie">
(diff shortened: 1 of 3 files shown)


ched88xtc (approved):
LGTM!)

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,103 @@
...
-
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
-        </div>
+import { findMovie } from '../../api/movies';
```
varseniuk, line 7: looks like there is no such function `findMovie` and file in your PR. From where do you import it? 🤔

varseniuk: Probably, you forgot to push all files?

varseniuk (changes requested):
https://user-images.githubusercontent.com/66570142/121026719-0dd1f580-c7af-11eb-99b0-3d6f7dd0578d.mov

We can add the same movie multiple times. Let's fix this
Also, why the outline of the input is red by default, like there's some error already?

varseniuk (approved):
👍

PR #233, opened 2021-06-04
Solution

ViPiven:
[DEMO LINK](https://ViPiven.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,98 @@
-import React from 'react';
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
 import './FindMovie.scss';
 
 import { MovieCard } from '../MovieCard';
-import movies from '../../api/movies.json';
+import { getMovie } from '../api/api';
 
-export const FindMovie = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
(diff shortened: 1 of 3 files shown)


ched88xtc (approved):

src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,98 @@
...
-
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
-        </div>
+import { getMovie } from '../api/api';
```
varseniuk, line 6: From where do you importing the function `getMovie`?
There's no file `api.js` in your project and no `gerMovie` function there

varseniuk: Seems like you haven't pushed all your files

varseniuk (changes requested):
https://user-images.githubusercontent.com/66570142/121022556-19bbb880-c7ab-11eb-91a6-db3f3dd91dc2.mov


We can search anything in your input and even if there's no movie found, we may add `nothing` to the list. Let's handle the error.
Also, why input has this red outline by default even without error?
And it seems like you haven't pushed all your files before PR

varseniuk (approved):
👍

PR #256, opened 2021-06-08
Solution

vitaliy-buchinchik:
[DEMO LINK](https://vitaliy-buchinchik.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.js
+++ b/src/components/FindMovie/FindMovie.js
@@ -1,55 +1,111 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import './FindMovie.scss';
 
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.js
```diff
@@ -1,55 +1,112 @@
...
-          >
-            Add to the list
-          </button>
+        <div className="field is-grouped">
+          <div className="control">
+            <button
+              type="button"
+              className="button is-light"
+              onClick={findMovie}
+            >
```
mykola-yarchuk, line 78: let me and other users press 'enter' and don't refresh the page, also it's cool if the user can just type the name of the movie and search for it with the `enter` button.

mykola-yarchuk (changes requested):
almost done

aliceblunt8 (approved):

PR #296, opened 2021-08-08
add to the list on submit

sergiy-fartushniak:
[demo](https://sergiy-fartushniak.github.io/react_movies-list-fetch-movies/)


VVVoloshina (approved):
Good)

danylo-onishchuk (changes requested):
If film is not found show an error message below the input. Hide the error after changing a title.

danylo-onishchuk (approved):

PR #325, opened 2021-09-17
Solution

sasha-morozov:
[DEMO LINK](https://sasha-morozov.github.io/react_movies-list-fetch-movies/)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,55 +1,116 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+// import classNames from 'classnames'
+import { getMovie } from '../../api/api';
```
DmytroMykoliv, line 4: Forgot to delete comments)

DmytroMykoliv (approved):
Looks good, don't see what else to add)

src/components/FindMovie/FindMovie.tsx
Elena-brv: remove comments

Elena-brv (approved):
the border shouldn't be red if there is no warning about not founded movie
image

PR #337, opened 2021-09-20
Movies list fetch movies

kristof22122:
[DEMO LINK](https://kristof22122.github.io/react_movies-list-fetch-movies/)


Elena-brv (changes requested):
nothing works
image

PR #339, opened 2021-09-20
Solution

poliandiya:
[DEMO LINK](https://poliandiya.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,4 @@
-import { Component } from 'react';
+import React from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -7,11 +7,20 @@ interface State {
   movies: Movie[];
 }
 
-export class App extends Component<{}, State> {
+export class App extends React.Component<{}, State> {
   state: State = {
     movies: [],
   };
```
Elena-brv, line 13: the task should be done using hooks

Elena-brv (changes requested):
the red border shouldn't be shown without warning
image

PR #340, opened 2021-09-20
movie list fetch movies by Eugene Gnedenko

set1one:
[DEMO LINK](https://set1one.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api/api.ts
+++ b/src/api/api.ts
@@ -1,0 +1,12 @@
+const API_URL = 'http://www.omdbapi.com/?apikey=4d03bd25';
+
+export const getMovie = (title: string): Promise<ApiMovie> => {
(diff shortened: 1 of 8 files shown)


src/api/api.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const getMovie = (title: string): Promise<ApiMovie> => {
+  return fetch(`${API_URL}&t=[${title}]`)
+    .then(response => {
+      if (!response.ok) {
+        throw new Error(`${response.status} - ${response.statusText}`);
+      }
+
+      return response.json();
+    });
```
danheim, line 11: use async await

danheim (changes requested):
Try to use async await, that's a good way

danheim (approved):

PR #346, opened 2021-09-20
Develop

Oleksandr-Zhitenko:
[DEMO LINK](https://Oleksandr-Zhitenko.github.io/react_movies-list-fetch-movies/)


Elena-brv (changes requested):
- nothing works
image


- warning shouldn't be shown when the page has just loaded and 'find a movie' button wasn't pressed
image

Elena-brv (changes requested):
- red border should be shown with a warning
- when I click 'find a movie' with an empty title the movie is founded 
image
- clear input after adding movie to the list

PR #350, opened 2021-09-21
Develop

RomSmile:
[DEMO LINK] https://RomSmile.github.io/react_movies-list-fetch-movies/


Elena-brv (changes requested):
- warning is shown all the time 
- empty movies shouldn't be founded and added to the list 

image

PR #455, opened 2022-01-17
Movie fetching done

ImIkator:
[Demo link](https://imikator.github.io/react_movies-list-fetch-movies/)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,81 @@
...
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
+              onChange={handleChange}
+            />
+          </div>
+
+          {!isMovieFound && (
+            <p className="help is-danger">
+              Can&apos;t find a movie with such a title
```
illiarerun, line 46: use template string here

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,81 @@
...
+import { MovieCard } from '../MovieCard';
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+type Props = {
+  searchMovie: (title: string) => void;
+  isMovieFound: boolean;
+  foundedMovie: Movie | null;
```
illiarerun, line 10:
```suggestion
  foundMovie: Movie | null;
```

illiarerun, line 10: Why do you need two of these? If movie is not found, the foundMovie var will have null value, thus meaning that movie hasn't been found
```suggestion
  foundedMovie: Movie | null;
```

src/App.tsx
```diff
@@ -1,29 +1,56 @@
...
 import { FindMovie } from './components/FindMovie';
+import { getMovie } from './api/getMovie';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [isMovieFound, setIsMovieFound] = useState(true);
+  const [foundMovie, setFoundMovie] = useState<Movie | null>(null);
```
illiarerun, line 12: so you don't need boolean value here also
```suggestion
  const [foundMovie, setFoundMovie] = useState<Movie | null>(null);
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,81 @@
...
+          <label className="label" htmlFor="movie-title">
+            Movie title
+          </label>
+
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
```
illiarerun, line 39: It is red all the time, the user will think that he made some type of mistake, fix it
image

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,81 @@
...
+
+          <div className="control">
+            <button
+              type="button"
+              className="button is-primary"
+              onClick={addToList}
+            >
+              Add to the list
+            </button>
+          </div>
```
illiarerun, line 71: It is active even whem movie hasn't been found yet, fix it
image

illiarerun (changes requested):
GJ 🔥 Now let's work on fixes:

PR #470, opened 2022-02-02
Solution

pasha-melnyk:
[DEMO LINK](https://pasha-melnyk.github.io/react_movies-list-fetch-movies/)


Elena-brv (changes requested):
good job, only add clearing input after adding a new movie

Elena-brv (approved):

PR #488, opened 2022-02-06
Develop

destroythemoon:
[DEMO LINK](https://destroythemoon.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,5 @@
+export function findMovie(title: string): Promise<Movie> {
+  return fetch(`https://www.omdbapi.com/?apikey=c87ce527&t=${title}`)
+    .then(response => response.json())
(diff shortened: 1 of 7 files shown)


src/api.ts
```diff
@@ -0,0 +1,5 @@
+export function findMovie(title: string): Promise<Movie> {
+  return fetch(`https://www.omdbapi.com/?apikey=c87ce527&t=${title}`)
+    .then(response => response.json())
+    .then(movie => movie);
```
Denchepornyuk, line 4: async await

Denchepornyuk (changes requested):
Screenshot 2022-02-14 at 17 45 49
There must be an error if I want to add a movie that is already in the list

PR #499, opened 2022-02-07
solution

OleksandrOstrovskyi:
[DEMO LINK](https://OleksandrOstrovskyi.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,56 @@
-import { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
+import { getMovie } from './api/api';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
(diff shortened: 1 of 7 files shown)


Elena-brv (changes requested):
- the movie shouldn't be found by every change in input. In this case, the button `find a movie` doesn't make sense
- error shouldn't be shown when the page was just loaded

src/App.tsx
```diff
@@ -1,29 +1,57 @@
...
+import { getMovie } from './api/api';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<(Movie | null)[]>([]);
+  const [input, setInput] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [clicked, setClicked] = useState(false);
```
Olexandr6, line 11: Move these to `FindMovie.tsx` component and them callback, as well.
This logic is connected with finding movies, so it shouldn't be here.

Olexandr6 (changes requested):
Check comment

PR #501, opened 2022-02-08
solution

Sesile26:
[Link](https://sesile26.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,100 @@
-import React from 'react';
+import { useState } from 'react';
 import './FindMovie.scss';
 
-// import { MovieCard } from '../MovieCard';
+import { MovieCard } from '../MovieCard';
 
-export const FindMovie: React.FC = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
-        <label className="label" htmlFor="movie-title">
-          Movie title
-        </label>
+async function getMovie(title: string): Promise<Movie> {
+  const res = await fetch(`https://www.omdbapi.com/?apikey=f151d5b2&t=${title}`);
 
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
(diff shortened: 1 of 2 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
+import { MovieCard } from '../MovieCard';
+
+async function getMovie(title: string): Promise<Movie> {
+  const res = await fetch(`https://www.omdbapi.com/?apikey=f151d5b2&t=${title}`);
+
+  return res.json();
+}
```
Elena-brv, line 10: move this function to the separate file

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [title, setTitle] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [movieWasFound, setMovieFound] = useState(true);
+
+  const handleTittleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```
Elena-brv, line 21:
```suggestion
  const handleTitleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
+    setTitle(event.target.value);
+    setMovieFound(true);
+  };
+
+  const foundMovie = async (titleNew: string) => {
```
Elena-brv, line 26: why do you need titleNew here? you already put it to the `title`

Elena-brv (changes requested):
image

- after adding new Movie or checking that movie is already exist preview should be cleared

PR #502, opened 2022-02-09

ViktorBachynskyi:


Denchepornyuk (changes requested):
Screenshot 2022-02-14 at 17 47 42
There must be an error if I want to add a movie that is already in the list

PR #503, opened 2022-02-09
solution

Viktor9r:
https://Viktor9r.github.io/react_movies-list-fetch-movies/


Elena-brv (changes requested):
- It shouldn't be possible to find a movie with an empty query 
image

- if I can't find a movie, I see the same strange preview
- after adding a movie to the list, preview should be cleared

PR #506, opened 2022-02-10
solution

gnosis-frog:
[DEMO LINK](https://gnosis-frog.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,101 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import { getMovie } from '../../api/movieDB';
+import { MovieCard } from '../MovieCard';
 
-// import { MovieCard } from '../MovieCard';
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
 
-export const FindMovie: React.FC = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
-        <label className="label" htmlFor="movie-title">
-          Movie title
-        </label>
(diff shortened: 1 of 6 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,101 @@
...
+      case 'False':
+        setSearchError(true);
+        break;
+
+      case 'True':
+        setSearchError(false);
+        break;
+
+      default:
+        break;
```
Elena-brv, line 28: no need to do switch/case when there are only 2 cases. `If` for one is enough

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,101 @@
...
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [movieTitle, setMovieTitle] = useState('');
+  const [searchError, setSearchError] = useState(false);
+
+  const findMovie = async (title: string) => {
```
Elena-brv, line 15: you can get title from state, no need to pass it as an argument

src/components/MovieCard/MovieCard.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import './MovieCard.scss';
 
 type Props = {
-  movie: Movie;
+  movie: Movie | null;
```
Elena-brv, line 5: you have already checked that movie is not null  in FindMovie component
```suggestion
  movie: Movie;
```

Elena-brv (changes requested):

PR #546, opened 2022-04-15
Develop

vitaliy-zhyvyliak:
[DEMO](https://vitaliy-zhyvyliak.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,6 @@
+const BASE_URL = 'http://www.omdbapi.com/?apikey=18d2cfad&t=';
+
+export const getData = (title: string) => {
+  return fetch(`${BASE_URL}${title}`)
+    .then(response => response.json());
+};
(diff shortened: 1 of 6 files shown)


src/api.ts
```diff
@@ -0,0 +1,6 @@
+const BASE_URL = 'https://www.omdbapi.com/?apikey=18d2cfad&t=';
```
Ihor-Karpyn, line 1:
```suggestion
const BASE_URL = 'https://www.omdbapi.com/?apikey=18d2cfad;
```

src/api.ts
```diff
@@ -0,0 +1,6 @@
...
+
+export const getData = (title: string) => {
+  return fetch(`${BASE_URL}${title}`)
+    .then(response => response.json());
+};
```
Ihor-Karpyn, line 6:
```suggestion
export const getData = (endpoint: string) => {
  return fetch(`${BASE_URL}${endpoint}`)
    .then(response => response.json());
};

export const getMovieByTitle = (title: string) => getDate(`&t=${title}`);
```

src/api.ts
Ihor-Karpyn, line 1:
```suggestion
const BASE_URL = 'https://www.omdbapi.com/?apikey=18d2cfad';
```

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #551, opened 2022-04-18
New commit

tumann21:
[DEMO LINK](https://tumann21.github.io/react_movies-list-fetch-movies/)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,118 @@
...
+              className="button is-light"
+              onClick={findMovie}
+            >
+              Find a movie
+            </button>
+          </div>
+
+          <div className="control">
+            <button
+              type="submit"
```
Bohdan-Maksymiuk, line 103: you may use type button and it will be no need to prevent default

Bohdan-Maksymiuk (approved):
Good job!

PR #558, opened 2022-05-04
Solution

k-romakovalchuk:
https://k-romakovalchuk.github.io/react_movies-list-fetch-movies/


aliceblunt8 (changes requested):
- input is red by default
Screenshot 2022-05-06 at 17 20 51

- you need to reset the form after adding the film to the list
Screenshot 2022-05-06 at 17 20 29

ArturErmolenko (changes requested):
Input still red and  doesn't reset the form after adding the film to the list, did you do deploy after last changes?

PR #562, opened 2022-05-18
Create react-movies-list-fetch-movies

Pavlo-Khashchevskyi:
[DEMO LINK](https://Pavlo-Khashchevskyi.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,37 @@
-import { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
+  // const [checkFilm, setCheckFilm] = useState(false);
 
-export class App extends Component<{}, State> {
-  state: State = {
-    movies: [],
+  const addMovie = (movie: Movie | undefined) => {
+    if (!movie) {
+      return;
+    }
+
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,29 +1,37 @@
...
+  // const [checkFilm, setCheckFilm] = useState(false);
 
-export class App extends Component<{}, State> {
-  state: State = {
-    movies: [],
-  };
+  const addMovie = (movie: Movie | undefined) => {
+    if (!movie) {
+      return;
+    }
```
Bohdan-Maksymiuk, line 13: do this check before calling `addMovie`, not inside

Bohdan-Maksymiuk (changes requested):
* remove film from preview after adding to the list

aliceblunt8 (changes requested):
preview is still not cleaned after adding a film

Screenshot 2022-05-19 at 16 32 43

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,110 @@
...
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [inputText, setInputText] = useState('');
+  const [previeData, setPrevieData] = useState<Movie | null>(null);
```
aliceblunt8, line 14:
```suggestion
  const [previewData, setPreviewData] = useState<Movie | null>(null);
```

aliceblunt8 (approved):

PR #608, opened 2022-06-14
solutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,87 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { request } from '../api/request';
+import { MovieCard } from '../MovieCard/MovieCard';
 import './FindMovie.scss';
 
-// import { MovieCard } from '../MovieCard';
+interface Props {
+  callbackForMovies: (arg0 : Movie[]) => void;
+  moviesPrev: Movie[];
+}
 
-export const FindMovie: React.FC = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
-        <label className="label" htmlFor="movie-title">
-          Movie title
-        </label>
+export const FindMovie: React.FC<Props> = (
+  { callbackForMovies, moviesPrev },
+) => {
+  const [movieForFind, setMovieForFind] = useState('');
+  const [preview, setPreview] = useState<Movie>();
 
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
-          />
+  async function finder() {
+    const result = await request(movieForFind.toLowerCase());
+
+    setPreview(result);
+  }
+
+  return (
+    <>
+      <form className="find-movie">
+        <div className="field">
+          <label className="label" htmlFor="movie-title">
+            Movie title
+          </label>
+
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
+              onChange={(event) => {
+                setMovieForFind(event.target.value);
+              }}
+            />
+          </div>
+
+          <p className="help is-danger">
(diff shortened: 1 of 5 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
+  const [isFilmValid, setIsFilmValid] = useState(true);
+
+  async function finder() {
+    const result = await request(movieForFind.toLowerCase());
+
+    if (result.Response !== 'False') {
+      setPreview(result);
+    } else {
+      setMovieForFind('');
+      setPreview(undefined);
```
KostiaO, line 26: It is better to set null instead of undefined

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
+            </button>
+          </div>
+
+          <div className="control">
+            <button
+              type="button"
+              className="button is-primary"
+              onClick={() => {
+                if (preview) {
+                  if (!moviesPrev.find(el => el.imdbID === preview.imdbID)) {
```
KostiaO, line 80: it is better to do this in App.tsx, so FindMovie component receive only function

KostiaO (changes requested):

src/components/FindMovie/FindMovie.tsx
EnjSadman: why?

Denchepornyuk (commented):
Screenshot 2022-06-20 at 13 57 14

the error should diaspore when we enter something at the input

Denchepornyuk:
Screenshot 2022-06-20 at 13 58 35

clear input only when `Add to the list`  button successfully added movie to list

src/components/FindMovie/FindMovie.tsx
Denchepornyuk: because of difference between null and undefined
also, undefined doesn't exist in json format

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie: React.FC<Props> = (
+  { callbackForMovies, moviesPrev },
+) => {
+  const [movieForFind, setMovieForFind] = useState('');
+  const [preview, setPreview] = useState<Movie>();
```
Denchepornyuk, line 16: preview can be null
```suggestion
  const [preview, setPreview] = useState<Movie | null>();
```

src/components/FindMovie/FindMovie.tsx
Denchepornyuk: agree, move this check to App.tsx component

Denchepornyuk, line 80: move all these logic to onSubmit method

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,100 @@
...
-        <div className="control">
-          <button
-            type="button"
-            className="button is-light"
-          >
-            Find a movie
-          </button>
+  return (
+    <>
+      <form className="find-movie">
```
Denchepornyuk, line 33: it  would be great to move form in a separate component

Denchepornyuk (changes requested):

src/components/api/Form/Form.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </div>
+
+      <div className="field is-grouped">
+        <div className="control">
+          <button
+            type="button"
+            className="button is-light"
+            onClick={() => {
+              finder();
+            }}
```
Denchepornyuk, line 82:
```
 onClick={finder}
```

Denchepornyuk (approved):

PR #613, opened 2022-06-17
Fetch movies was done

vlasiuk-anatolii:
[DEMO LINK](https://vlasiuk-anatolii.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,0 +1,7 @@
+const BASE_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=257b4347&t';
+
+export const getMovie = async (title: string) => {
+  const response = await fetch(`${BASE_URL}=[${title}]`);
+
+  return response.json();
+};
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,102 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import { MovieCard } from '../MovieCard';
+import { getMovie } from '../../api';
 
-// import { MovieCard } from '../MovieCard';
+type Props = {
+  addMovies: (film: Movie) => void;
+};
 
-export const FindMovie: React.FC = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
-        <label className="label" htmlFor="movie-title">
-          Movie title
-        </label>
+export const FindMovie: React.FC<Props> = ({ addMovies }) => {
+  const [movie, setMovie] = useState(null);
+  const [query, setQuery] = useState('');
+  const [error, setError] = useState('');
 
-        <div className="control">
-          <input
-            type="text"
-            id="movie-title"
-            placeholder="Enter a title to search"
-            className="input is-danger"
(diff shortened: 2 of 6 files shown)


src/api.ts
```diff
@@ -0,0 +1,7 @@
+const BASE_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=257b4347&t';
```
AnastasiyaSopotnytska, line 1: Send a request to https://www.omdbapi.com/?apikey=[yourkey]&t=[title]

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,102 @@
...
+          <label className="label" htmlFor="movie-title">
+            Movie title
+          </label>
+
+          <div className="control">
+            <input
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className="input is-danger"
```
AnastasiyaSopotnytska, line 58: Add class 'is-danger' only if film is not found

AnastasiyaSopotnytska (approved):

Denchepornyuk:
Screenshot 2022-06-20 at 14 59 17

something wrong with input with error

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,106 @@
...
+type Props = {
+  addMovies: (film: Movie) => void;
+};
 
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie: React.FC<Props> = ({ addMovies }) => {
+  const [movie, setMovie] = useState(null);
```
Denchepornyuk, line 11: add type for movie

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,106 @@
...
-            className="button is-light"
-          >
-            Find a movie
-          </button>
+    setError(false);
+  };
+
+  return (
+    <>
+      <form
```
Denchepornyuk, line 47: would be great to move form in separate component

Denchepornyuk (changes requested):
good job

vlasiuk-anatolii:
> 



> something wrong with input when we have error

I don't understend what wrong with it. I tested it a long time.

src/components/FindMovie/FindMovie.tsx
vlasiuk-anatolii: > would be great to move form in separate component

Would you explain me why we should do this?  Form possesses all elements of FindeMovie.

Denchepornyuk:
> > 
> 
> > something wrong with input when we have error
> 
> I don't understend what wrong with it. I tested it a long time.


Now It works as expected

src/components/FindMovie/FindMovie.tsx
Denchepornyuk: Because FindMovie has a lot of logic and it's a large component so it's better to split some login between components and make FindMovie more clear and readable

Denchepornyuk (approved):

PR #616, opened 2022-06-17
solution

ZUBOV-ILLIA:
[DEMO LINK](https://ZUBOV-ILLIA.github.io/react_movies-list-fetch-movies/)


yelnikov-andrii (approved):

denys-cheporniuk (changes requested):

ZUBOV-ILLIA:
> 

I did it intentionally, because I thought that if the field is empty, then there is no point in displaying an error)
Fixed
image

denys-cheporniuk (approved):

PR #618, opened 2022-06-18
solution

SavMary:
[DEMO LINK](https://savmary.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,106 @@
-import React from 'react';
+import React, { useState, FormEvent } from 'react';
+import classNames from 'classnames';
+import { getMovie } from '../../api/api';
 import './FindMovie.scss';
 
-// import { MovieCard } from '../MovieCard';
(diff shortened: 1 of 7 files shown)


vlad-demchuk (approved):

vlad-demchuk (changes requested):
demo link doesn't work

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,106 @@
...
-        <p className="help is-danger">
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+type Props = {
+  onAddMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ onAddMovie }) => {
+  const [movie, setMovie] = useState<Movie>();
```
Denchepornyuk, line 13: movie can be null 
```suggestion
  const [movie, setMovie] = useState<Movie | null>(null);
```

Denchepornyuk (approved):

PR #631, opened 2022-06-23
Solution

v-vytrykush:
[DEMO](https://v-vytrykush.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/API/api.ts
+++ b/src/API/api.ts
@@ -1,0 +1,8 @@
+const API_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=9d3a21ab';
+
+export const getFilmByTitle = (title: string): Promise<Movie> => {
+  return fetch(`${API_URL}&t=${title}`)
+    .then(response => {
+      return response.json();
+    });
+};
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,19 @@
-import { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
 
-export class App extends Component<{}, State> {
-  state: State = {
-    movies: [],
-  };
-
-  render() {
-    const { movies } = this.state;
-
-    return (
-      <div className="page">
-        <div className="page-content">
-          <MoviesList movies={movies} />
-        </div>
-        <div className="sidebar">
-          <FindMovie />
-        </div>
+  return (
+    <div className="page">
+      <div className="page-content">
+        <MoviesList movies={movies} />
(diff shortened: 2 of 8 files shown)


src/API/api.ts
```diff
@@ -0,0 +1,8 @@
+const API_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=9d3a21ab';
```
AndriiRut, line 1: You are using default link which was send to you after registration. default link already get you on the one and only movie  Try to find out by yourself how to use the api-link correctly. Couse in link the part `i=tt389619` means id of movie and `key=9d3a21ab` was send to you separately in the same massage as the link.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,128 @@
...
+          return setPreview(film);
+        }
+
+        return null;
+      });
+  };
+
+  const addmovie = () => {
+    if (preview
+      && movies.every(movie => movie.imdbID !== preview.imdbID)) {
```
AndriiRut, line 33: I would use `.findindex` or `.indexOf` instead of `.every`. Couse `.every` will do iteration to the end of the array but the object which you try to find in the midle /start.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,128 @@
...
-          Can&apos;t find a movie with such a title
-        </p>
-      </div>
+export const FindMovie: React.FC<Props> = ({ selectedMovie, movies }) => {
+  const [title, setTitle] = useState('');
+  const [preview, setPreview] = useState<Movie | null>(null);
+  const [error, setError] = useState(false);
+  const [duplicate, setDuplicate] = useState(false);
+
+  const findMovie = () => {
```
AndriiRut, line 18: I don't shure, but didn't JS prevent to name variables/functions with already used names in the same component? You have used `findMovie `  2 times.

src/App.tsx
```diff
@@ -1,29 +1,19 @@
...
-        <div className="page-content">
-          <MoviesList movies={movies} />
-        </div>
-        <div className="sidebar">
-          <FindMovie />
-        </div>
+  return (
+    <div className="page">
+      <div className="page-content">
+        <MoviesList movies={movies} />
```
AndriiRut, line 12: I don't see `id` in the props of `MoviesList` component. are you used attribute `key` while maping the list?

AndriiRut (changes requested):
Good work! I impressed by your user friendly interface of serching (sc. the Errors and massages to users). But still you have wat to improve!

src/components/FindMovie/FindMovie.tsx
v-vytrykush: the syntax is differend

AndriiRut: oh I see! yeh, first later is capital in component!

Anna-Tkachenko (approved):
Well Done!

PR #632, opened 2022-06-24
My solve

Yaroslav-Kononenko:
[DEMO LINK](https://Yaroslav-Kononenko.github.io/react_movies-list-fetch-movies/)


Anna-Tkachenko (changes requested):
i can't find the movie
image

PR #636, opened 2022-06-25
First_commit

igoree88:
[DEMO LINK](https://igoree88.github.io/react_movies-list-fetch-movies/)


Anna-Tkachenko (changes requested):
i can't find any movie after this error
image

PR #649, opened 2022-07-05
Fetch movies

proznerka:
[DEMO LINK](https://proznerka.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,29 +1,28 @@
-import { Component } from 'react';
+import React, { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 
-interface State {
-  movies: Movie[];
-}
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
 
-export class App extends Component<{}, State> {
-  state: State = {
-    movies: [],
+  const addMovie = (movie: Movie) => {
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,29 +1,28 @@
...
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
 
-export class App extends Component<{}, State> {
-  state: State = {
-    movies: [],
-  };
+  const addMovie = (movie: Movie) => {
+    const isAddedBefore = movies
+      .some(mov => mov.imdbID === movie.imdbID);
```
ihor-jpeg, line 11: `mov`, because `movie` is too long? 😆

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,107 @@
...
+  const [currentMovie, setCurrentMovie] = useState<Movie | null>(null);
+  const [searchError, setSearchError] = useState(false);
+
+  const handleSearch = async () => {
+    if (title) {
+      const movieFromServer = await getMovie(title);
+
+      setSearchError(movieFromServer.Response === 'False');
+
+      setCurrentMovie(movieFromServer);
```
ihor-jpeg, line 23: maybe if there is an error we should not set current movie?

ihor-jpeg (changes requested):

PR #653, opened 2022-07-10
Task complete

rHashTag:
[DEMO LINK](https://rhashtag.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,54 +1,102 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
 import './FindMovie.scss';
 
-// import { MovieCard } from '../MovieCard';
+import { getMovie } from '../../api';
+import { MovieCard } from '../MovieCard';
 
-export const FindMovie: React.FC = () => (
-  <>
-    <form className="find-movie">
-      <div className="field">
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,102 @@
...
+          </label>
+
+          <div className="control">
+            <input
+              value={title}
+              onChange={handleInput}
+              type="text"
+              id="movie-title"
+              placeholder="Enter a title to search"
+              className={cn('input', { 'is-danger': !title })}
```
iLebo-stack, line 55: I think you should use `searchError` instead of `!title` because you input has red borders right from the start and after adding the movie to the list

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,54 +1,102 @@
...
+    if (currentMovie) {
+      addMovie(currentMovie);
+      setTitle('');
+      setCurrentMovie(null);
+    }
+  };
+
+  return (
+    <>
+      <form className="find-movie">
```
iLebo-stack, line 42: You should add `event.preventDefault()` to prevent default behaviour on your form when submitting because your app reloads when pressing enter from the input

iLebo-stack (changes requested):
Good job👍, check my comments

PR #678, opened 2022-07-28
Created Movies List with Form

olya-yevsieienko:
[DEMO LINK](https://olya-yevsieienko.github.io/react_movies-list-fetch-movies/)


vsupruniuk (approved):
Nice🔥

mykola-koval (approved):
would be good if you show an error when the movie is already on the list

PR #687, opened 2022-07-28
Add fetch movies

ghost:
[DEMO LINK](https://juajum.github.io/react_movies-list-fetch-movies/)


viktorlyt (approved):
Well done!
Only several remarks from me:
1. "Preview" - better not to be shown when no movie found
2. What about a spinner ?

juajum:
> Well done! Only several remarks from me:
> 
> 1. "Preview" - better not to be shown when no movie found
> 2. What about a spinner ?

good points, thanks, I'll fix it😊

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,28 +1,80 @@
...
+              }}
             />
           </div>
 
-          <p className="help is-danger" data-cy="errorMessage">
-            Can&apos;t find a movie with such a title
-          </p>
+          {error && (
+            <p className="help is-danger" data-cy="errorMessage">
+              Can&apos;t find a movie with such a title
```
jstmpelowycz, line 75: you'd rather use interpolation

jstmpelowycz (changes requested):

PR #708, opened 2022-08-02
solution was added

Bielichenko:
[DEMO LINK](https://Bielichenko.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,53 @@
-import React from 'react';
-import './FindMovie.scss';
+/* eslint-disable no-console */
+import React, { useState } from 'react';
+import classNames from 'classnames';
 
-export const FindMovie: React.FC = () => {
+import { getMovie } from '../../api';
+import './FindMovie.scss';
+import { MovieData } from '../../types/MovieData';
+import { MovieCard } from '../MovieCard';
+
+type Props = {
+  setSelectedMovie: (movie: MovieData) => void;
+};
+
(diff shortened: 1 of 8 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,53 @@
...
+
+type Props = {
+  setSelectedMovie: (movie: MovieData) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ setSelectedMovie }) => {
+  const [input, setInput] = useState('');
+  const [isLoading, setIsLoading] = useState(false);
+  const [movie, setMovie] = useState<MovieData | null>(null);
+  const [error, setError] = useState<boolean>(false);
```
jstmpelowycz, line 18:
```suggestion
  const [error, setError] = useState(false);
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +63,65 @@ export const FindMovie: React.FC = () => {
               id="movie-title"
               placeholder="Enter a title to search"
               className="input is-dander"
+              value={input}
+              onChange={(event) => setInput(event.target.value)}
             />
           </div>
+          {error && (
+            <p className="help is-danger" data-cy="errorMessage">
+              Can&apos;t find a movie with such a title
```
jstmpelowycz, line 72: use interpolation

jstmpelowycz (changes requested):
there is an issue – when query changed (user starts typing), error does not disappear, you should fix it

PR #710, opened 2022-08-06
Solution with "Enter" press feature

andriy-fesych:
[DEMO LINK](https://andriy-fesych.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -13,7 +13,7 @@ export const FindMovie: React.FC<Props> = ({ addMovie }) => {
   const [loading, setLoading] = useState<boolean>(false);
   const [error, setError] = useState<boolean>(false);
   const [movieFound, setMovieFound]
-    = useState<Movie | null>();
+    = useState<Movie | null>(null);
   const [query, setQuery] = useState<string>('');
 
   const loadMovie = () => {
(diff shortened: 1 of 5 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,70 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import classNames from 'classnames';
+import { MovieCard } from '../MovieCard';
+import { Movie } from '../../types/Movie';
+import { getMovie } from '../../api';
+
+type Props = {
+  addMovie: (arg0: Movie) => void | null
```
alexandra-protyanova, line 9: not the best name for variable

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,70 @@
...
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [loading, setLoading] = useState<boolean>(false);
```
alexandra-protyanova, line 13: it is not necessary for primitive types
```suggestion
  const [loading, setLoading] = useState(false);
```

alexandra-protyanova (changes requested):
It would be great to show any message, when user trying to add existing movie in the list

alexandra-protyanova (approved):

PR #712, opened 2022-08-08
react movies list

IvanG6:
https://ivang6.github.io/react_movies-list-fetch-movies/

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,46 @@
-import React from 'react';
+import React, { useState } from 'react';
+import cn from 'classnames';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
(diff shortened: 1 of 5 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,46 @@
...
+  const [hasError, setHasError] = useState(false);
+
+  const searchHandler = async () => {
+    setIsLoading(true);
+
+    try {
+      const res = await getMovie(searchTerm);
+
+      if ('Title' in res) {
+        const newMovie: Movie = {
```
alexandra-protyanova, line 25: it is redundant to create object, res is exactly what u need

alexandra-protyanova (changes requested):
Let's show the message when user trying to add film, which is already in list, not just will not do it)

ArturErmolenko (approved):

PR #734, opened 2022-08-23
Solution

Sergii2705:
- [DEMO LINK](https://Sergii2705.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,11 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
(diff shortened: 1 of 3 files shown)


Ihor-Karpyn (changes requested):
image

Sergii2705:
> image

This is not my fault, everything works for me. This error is initiated by the browser and appears due to the fact that a secure page (loaded via https) cannot use resources (scripts, styles) from insecure sources (i.e. those that work via http).
Therefore, the browser blocks the download of such resources.
I changed the http to https and it should work now.

ivan-kadykalo (approved):
🔥

PR #744, opened 2022-08-24
Movie fetch

Yura-Lysenko:


Ihor-Karpyn (approved):
image


add btn should be disabled if there isn't film

PR #755, opened 2022-08-26
Solution

alexbardyshev:
[DEMO LINK](https://alexbardyshev.github.io/react_movies-list-fetch-movies/)


Ihor-Karpyn (approved):
clear input after added film to the list

PR #778, opened 2022-09-20
Solution (react_movies-list-fetch-movies)

nosovandriy:
[DEMO LINK](https://nosovandriy.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,61 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import classNames from 'classnames';
 
-export const FindMovie: React.FC = () => {
+import { MovieData } from '../../types/MovieData';
+import { getMovie } from '../../api';
+import { MovieCard } from '../MovieCard/MovieCard';
+
+type Props = {
+  addNewMovie: (newMovie: MovieData | null) => void,
+};
+
+export const FindMovie: React.FC<Props> = ({ addNewMovie }) => {
+  const [movie, setMovie] = useState<MovieData | null>(null);
+  const [movieTitleQuery, setMovieTitleQuery] = useState('');
+  const [hasLoadingError, setHasLoadingError] = useState(false);
+  const [loadingMovie, setLoadingMovie] = useState(false);
+
+  const handleLoadMovieFromServer = async () => {
+    setLoadingMovie(true);
+
+    try {
+      const newMovie = await getMovie(movieTitleQuery);
+
+      if (!newMovie.imdbID) {
+        setMovie(null);
+        setHasLoadingError(true);
+      } else {
+        setMovie(newMovie);
+        setHasLoadingError(false);
+      }
+    } catch (error) {
+      alert('Error loading data from the server');
+    } finally {
+      setLoadingMovie(false);
+    }
+  };
+
+  const handleFindMovie = (query: string) => {
+    if (query !== movieTitleQuery) {
+      setHasLoadingError(false);
+    }
+
+    setMovieTitleQuery(query);
+  };
+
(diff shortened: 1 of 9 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +68,60 @@ export const FindMovie: React.FC = () => {
...
-              className="button is-primary"
-            >
-              Add to the list
-            </button>
+            {movie && (
+              <button
+                data-cy="addButton"
+                type="button"
+                className="button is-primary"
+                onClick={handleAddNewMovieToTheList}
```
SanyaBratashchuk, line 110: do it on submit

SanyaBratashchuk (changes requested):

src/components/FindMovie/FindMovie.tsx
nosovandriy: I changed it, but I don't know if I correctly understood what I needed to do.
Can you explain, why my solution was wrong and why your solution is better?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +69,59 @@ export const FindMovie: React.FC = () => {
...
+              className={classNames(
+                'button is-light',
+                { 'is-loading': loadingMovie },
+              )}
+              onClick={handleLoadMovieFromServer}
             >
-              Find a movie
+              {!movie
+                ? 'Find a movie'
+                : 'Search again'}
```
Bohdan-Maksymiuk, line 100: prepare this text before return (in this way you markup will be more clear)

src/components/MovieCard/MovieCard.tsx
```diff
@@ -1,44 +1,55 @@
...
+
+  return (
+    <div className="card" data-cy="movieCard">
+      <div className="card-image">
+        <figure className="image is-4by3">
+          <img
+            data-cy="moviePoster"
+            src={Poster === 'N/A'
+              ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+              : Poster}
```
Bohdan-Maksymiuk, line 25: same here

Bohdan-Maksymiuk (approved):
Good job 👍

src/components/FindMovie/FindMovie.tsx
Bohdan-Maksymiuk: it wasn't wrong if logic is correct, but will be more semantic and logically  correct to use onSubmit handler if you use form in your solution

src/components/MovieCard/MovieCard.tsx
```diff
@@ -1,44 +1,57 @@
...
-        </div>
+export const MovieCard: React.FC<Props> = ({ movie }) => {
+  const {
+    Poster,
+    Title,
+    Plot,
+    imdbID,
+  } = movie;
+
+  const changePoster = (Poster === 'N/A')
```
Bohdan-Maksymiuk, line 17:
```suggestion
  const posterSrc = (Poster === 'N/A')
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,64 @@
...
+  };
+
+  const handleAddNewMovieToTheList = (eventSubmit: React.FormEvent) => {
+    eventSubmit.preventDefault();
+    addNewMovie(movie);
+    setMovie(null);
+    setMovieTitleQuery('');
+  };
+
+  const changeFindButton = !movie ? 'Find a movie' : 'Search again';
```
Bohdan-Maksymiuk, line 54:
```suggestion
  const findButtonText = !movie ? 'Find a movie' : 'Search again';
```

Bohdan-Maksymiuk (approved):

PR #790, opened 2022-09-27
Implement fetch movies

Desertiks:
[DEMO LINK](https://Desertiks.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,7 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
 
   return (
     <div className="page">
@@ -14,7 +14,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie setMovies={setMovies} />
       </div>
     </div>
   );
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,7 +1,7 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=d95d1d85';
 
 export function getMovie(query: string): Promise<MovieData | ResponseError> {
   return fetch(`${API_URL}&t=${query}`)
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,85 @@
-import React from 'react';
+import classNames from 'classnames';
+import React, {
+  ChangeEvent, Dispatch, FormEvent, SetStateAction, useState,
+} from 'react';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
 import './FindMovie.scss';
 
-export const FindMovie: React.FC = () => {
+type Props = {
+  setMovies: Dispatch<SetStateAction<Movie[]>>,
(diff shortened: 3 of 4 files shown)


src/App.tsx
```diff
@@ -14,7 +14,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie setMovies={setMovies} />
```
dellachaise, line 17: Better create an event handler and pass it to props than just `setMovies`.

src/api.ts
```diff
@@ -1,7 +1,7 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=d95d1d85';
```
dellachaise, line 4: Better use a variable for `key`.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,85 @@
...
+    const data = await getMovie(query);
+
+    if (!('Error' in data)) {
+      setError(false);
+
+      const {
+        Plot, Poster, Title, imdbID,
+      } = data;
+
+      // eslint-disable-next-line
```
dellachaise, line 35: Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find the way to solve it without `eslint-disable`.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,85 @@
...
+        imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+      });
+    } else {
+      setError(true);
+    }
+
+    setIsLoading(false);
+  };
+
+  const handleMovieQuery = (event: ChangeEvent<HTMLInputElement>) => {
```
dellachaise, line 52: Methods instead of inline event handlers 👍

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,85 @@
...
+    setError(false);
+    setQuery(event.target.value);
+    setSameMovie(false);
+  };
+
+  const handleAddMovie = () => {
+    if (movie) {
+      setMovies((currentMovies) => {
```
dellachaise, line 60: Functional set state 🔥  Just move this logic to `App.tsx`.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,85 @@
...
+    setMovie(null);
+    setQuery('');
+    setClicked(false);
+    setError(false);
+  };
+
   return (
     <>
-      <form className="find-movie">
+      <form onSubmit={(event) => handleFindMovie(event)} className="find-movie">
```
dellachaise, line 82: `onSubmit` is quite a long argument, better rewrite it this way 👇  It would be easier to read.
```suggestion
      <form 
        onSubmit={(event) => handleFindMovie(event)} 
        className="find-movie"
      >
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +92,70 @@ export const FindMovie: React.FC = () => {
...
-            <button
-              data-cy="addButton"
-              type="button"
-              className="button is-primary"
+              className={classNames(
+                'button',
+                'is-light',
+                {
+                  'is-loading': isLoading,
+                },
```
dellachaise, line 126: If line length allows - better always write this way 👇 Code would be more readable.
```suggestion
                { 'is-loading': isLoading },
```

dellachaise (changes requested):
Good job 👍 

Just need to add some improvements according to the comments. And pay attention to them when you would write your future test tasks. I also marked some things that you did great 👍  (keep do it 🔥 )

PR #796, opened 2022-10-04
Solution

Black3942:


src/App.tsx
```diff
@@ -5,7 +5,13 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie) => {
+    if (movies.every(element => element.imdbId !== movie.imdbId)) {
```
alexandra-protyanova, line 11: I think, it would be great to show message for user why don't u add current movie to the list)

alexandra-protyanova (approved):

PR #800, opened 2022-10-09
Task done

sanyokmalyshev:
[DEMO LINK](https://sanyokmalyshev.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,62 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import { getMovie } from '../../api';
+import { MovieCard } from '../MovieCard';
+import { Movie } from '../../types/Movie';
 
-export const FindMovie: React.FC = () => {
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [querySearch, setQuerySearch] = useState('');
+  const [movie, setMovie] = useState<Movie>();
+  const [error, setError] = useState(false);
+  const [isLoading, setLoading] = useState(false);
+
+  const handleClick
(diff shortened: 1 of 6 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,62 @@
...
+import { Movie } from '../../types/Movie';
+
+type Props = {
+  addMovie: (movie: Movie) => void;
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [querySearch, setQuerySearch] = useState('');
+  const [movie, setMovie] = useState<Movie>();
+  const [error, setError] = useState(false);
```
polosanya, line 14: For boolean values it is better to give names starting with is/are/can etc. so it's clear that value has boolean type.
image

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,62 @@
...
+      e.preventDefault();
+      setLoading(true);
+      const result = await getMovie(querySearch);
+
+      setLoading(false);
+
+      if ('Title' in result) {
+        const {
+          Title, Plot, Poster, imdbID,
+        } = result;
```
polosanya, line 28:
```suggestion
      const response = await getMovie(querySearch);

      setLoading(false);

      if ('Title' in response) {
        const {
          Title, Plot, Poster, imdbID,
        } = result;
```
It is better to name it `response`, not result. Because response can be also an error.
Also I think it is better to check if response has `Error`, not if response doesn't have `Title`. Would be more logical

polosanya (changes requested):
Good job! Just please check my comments. I think they can improve your code a little and make it more clear)

src/components/FindMovie/FindMovie.tsx
sanyokmalyshev: Thanks, done

polosanya (approved):
Well done!

PR #803, opened 2022-10-15
implemented film search

ihorandrianov:
[DEMO LINK](https://ihorandrianov.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,27 +1,25 @@
-import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 import { Movie } from './types/Movie';
 
 const API_URL = 'https://www.omdbapi.com/?apikey=dd901605';
 
 export async function getMovie(query: string): Promise<Movie | ResponseError> {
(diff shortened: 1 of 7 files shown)


src/api.ts
```diff
@@ -1,13 +1,27 @@
...
-export function getMovie(query: string): Promise<MovieData | ResponseError> {
-  return fetch(`${API_URL}&t=${query}`)
-    .then(res => res.json())
-    .catch(() => ({
-      Response: 'False',
-      Error: 'unexpected error',
-    }));
+export async function getMovie(query: string): Promise<Movie | ResponseError> {
+  const response: MovieData | ResponseError = await fetch(`${API_URL}&t=${query}`)
+    .then(res => res.json());
```
Denchepornyuk, line 9: don't use `then` with `await`, use await in all places, avoid `then`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #816, opened 2022-11-01
Implemented the movies list with API search

anastasiia-nahaiska:
[DEMO LINK](https://anastasiia-nahaiska.github.io/react_movies-list-fetch-movies/)


jojnajbude (approved):
Good job, but implement if tried to add film which already in movies list.

anastasiia-nahaiska:
> Good job, but implement if tried to add film which already in movies list.

Thanks) But it works ([video](https://www.loom.com/share/bc7367ace5ad44ae8dec666b2e4effb6))
Do you mean something else?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,81 @@
...
+
+    try {
+      const movieFromApi = await getMovie(query);
+
+      if ('Error' in movieFromApi) {
+        setHasError(true);
+        setLoading(false);
+
+        return;
+      }
```
Anna-Tkachenko, line 54: can we throw the error in the api function and catch the error in the catch {} structure?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,81 @@
...
+        title: Title,
+        description: Plot,
+        imgUrl: Poster === 'N/A'
+          ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+          : Poster,
+        imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+        imdbId: imdbID,
+      });
+    } catch (error) {
+      setHasError(true);
```
Anna-Tkachenko, line 73: should you reset loading variable here?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,81 @@
...
+
+      setFindedMovie({
+        title: Title,
+        description: Plot,
+        imgUrl: Poster === 'N/A'
+          ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+          : Poster,
+        imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+        imdbId: imdbID,
+      });
```
Anna-Tkachenko, line 71: let's use separate variable where we gather all info about new object, and pass it here

Anna-Tkachenko (changes requested):

src/components/FindMovie/FindMovie.tsx
anastasiia-nahaiska: I hope I understood correctly ^-^

Denchepornyuk: agree, it will be better solution

Denchepornyuk (approved):

PR #817, opened 2022-11-02
Solution

TheVovchik:
**[DEMO LINK](https://thevovchik.github.io/react_movies-list-fetch-movies/)**

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,13 +1,13 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=378681db';
 
 export function getMovie(query: string): Promise<MovieData | ResponseError> {
   return fetch(`${API_URL}&t=${query}`)
     .then(res => res.json())
-    .catch(() => ({
+    .catch((error) => ({
       Response: 'False',
-      Error: 'unexpected error',
+      Error: error,
     }));
 }
--- a/src/components/AppProvider/AppProvider.tsx
+++ b/src/components/AppProvider/AppProvider.tsx
@@ -1,30 +1,32 @@
 import React, { useState } from 'react';
+import { getMovie } from '../../api';
 import { Movie } from '../../types/Movie';
+import { MovieData } from '../../types/MovieData';
 
 type Context = {
   movies: Movie[],
-  setMovies: (newMovies: Movie[]) => void,
   query: string,
-  setQuery: (newQuery: string) => void,
   movie: Movie | null,
-  setMovie: (newMovie: Movie | null) => void,
   isMovieExist: boolean,
-  setIsMovieExist: (newCheck: boolean) => void,
   isSearch: boolean,
-  setIsSearch: (newStatus: boolean) => void,
+
+  handleInput: (event: React.ChangeEvent<HTMLInputElement>) => void,
+  getMovieFromServer: (appliedQuery: string) => void,
+  handleSearch: (event: React.FormEvent<HTMLFormElement>) => void,
+  addingMovie: () => void,
 };
 
 export const AppContext = React.createContext<Context>({
   movies: [],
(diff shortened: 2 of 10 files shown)


ihorandrianov:
Add warning when trying to add duplicate movie!

src/api.ts
```diff
@@ -1,13 +1,13 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=378681db';
 
 export function getMovie(query: string): Promise<MovieData | ResponseError> {
```
ihorandrianov, line 6: Some mentors asks to rewrite this function with async syntax

src/components/Button/Button.tsx
```diff
@@ -0,0 +1,16 @@
+import { FC } from 'react';
+
+type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  children: React.ReactNode;
+};
+
+export const Button: FC<Props> = ({ children, ...props }) => {
```
ihorandrianov, line 7: Why you need Button wrapper?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,95 @@
...
+  const checkMovie = (newMovie: Movie) => {
+    return movies.some(film => film.imdbId === newMovie.imdbId);
+  };
+
+  const getMovieFromServer = async (appliedQuery: string) => {
+    const newFilm = await getMovie(appliedQuery);
+
+    setIsSearch(false);
+
+    if (Object.prototype.hasOwnProperty.call(newFilm, 'Poster')) {
```
ihorandrianov, line 44: You can simplify this syntax

src/components/AppProvider/AppProvider.tsx
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const AppProvider: React.FC = ({ children }) => {
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [isMovieExist, setIsMovieExist] = useState(false);
+  const [isSearch, setIsSearch] = useState(false);
+
+  const contextValue = useMemo(() => ({
```
ihorandrianov, line 37: Using useMemo without good reason is bad practice, overall this is strange use case to memoize function that did not calculate anything and just returns object of const values

src/components/Button/Button.tsx
TheVovchik: simple practice to move some common logic

ihorandrianov (approved):

src/api.ts
```diff
@@ -1,13 +1,13 @@
...
   return fetch(`${API_URL}&t=${query}`)
     .then(res => res.json())
-    .catch(() => ({
+    .catch((error) => ({
       Response: 'False',
-      Error: 'unexpected error',
+      Error: error,
```
Anna-Tkachenko, line 11: you can also throw new error here and handle this error in catch {} while calling api function

src/components/AppProvider/AppProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+  setIsMovieExist: (newCheck: boolean) => void,
+  isSearch: boolean,
+  setIsSearch: (newStatus: boolean) => void,
+};
+
+export const AppContext = React.createContext<Context>({
+  movies: [],
+  setMovies: () => {},
+  query: '',
+  setQuery: () => {},
```
Anna-Tkachenko, line 21: you shouldn't add variables that can be changes on each click to the context
use context only for general values for all App, that can affect on many components and will use everywhere

src/components/Button/Button.tsx
Anna-Tkachenko: good job with separate component for buttons!

Anna-Tkachenko (approved):
Good job!

PR #820, opened 2022-11-02
Implemented movies list with add option from server

Andrew-Skoryk:
[DEMO LINK](https://Andrew-Skory.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,90 @@
-import React from 'react';
+import {
+  FC,
+  FormEvent,
+  useMemo,
+  useState,
+  memo,
+  ChangeEvent,
+} from 'react';
+import cn from 'classnames';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
 import './FindMovie.scss';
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,91 @@
...
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [isIncluded, setIsIncluded] = useState(false);
+
+  // eslint-disable-next-line max-len
+  const DEAFULT_PICTURE = useMemo(() => (
+    'https://via.placeholder.com/360x270.png?text=no%20preview'
+  ), []);
```
Anna-Tkachenko, line 30: will be better to move constants outside of the component

Anna-Tkachenko: and you shouldn't cover primitive data types with useMemo
we need useMemo for functions / objects because for each re-rendering corresponding variables will store the new link for some function or object
primitive values doesn't change in this way

Anna-Tkachenko (approved):

PR #821, opened 2022-11-03
Solution

MihailKostov:
[DEMO LINK](https://MihailKostov.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,18 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (newMovie: Movie) => {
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -5,7 +5,18 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (newMovie: Movie) => {
+    if (movies.find(movie => movie.imdbId === newMovie.imdbId)) {
```
AlisaVlasova, line 11:
```suggestion
    if (movies.some(movie => movie.imdbId === newMovie.imdbId)) {
```
`find` method returns you whole object, if you need here only boolean it will be better to use `some`

AlisaVlasova (changes requested):

nazarmatsevych (approved):

PR #823, opened 2022-11-04
find movies fetch implemented

artem-musii:
[DEMO LINK](https://artem-musii.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,15 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie) => {
+    if (movies.find(movieInList => movieInList.imdbId === movie.imdbId)) {
+      setMovies(movies);
+    } else {
+      setMovies(currentMovies => [...currentMovies, movie]);
+    }
+  };
 
   return (
     <div className="page">
@@ -14,7 +22,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie addMovie={addMovie} />
       </div>
     </div>
   );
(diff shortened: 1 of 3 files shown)


iPanda177 (approved):
image


Cool!)
One small thing to fix)
i haven't started the search yet and the button already says "search again" :)

src/App.tsx
```diff
@@ -5,7 +5,15 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie) => {
+    if (movies.find(movieInList => movieInList.imdbId === movie.imdbId)) {
```
Anna-Tkachenko, line 11: will be better to store found movie in separate variable instead of call find method inside the if

src/api.ts
```diff
@@ -1,13 +1,19 @@
...
+  const response = await fetch(`${API_URL}&t=${query}`);
+
+  try {
+    return await response.json();
+  } catch {
+    return ({
       Response: 'False',
       Error: 'unexpected error',
-    }));
+    });
```
Anna-Tkachenko, line 17: use throw new Error instead of returning such object
in this way you'll be available to handle this error in catch {} while calling this function

Anna-Tkachenko (changes requested):

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,143 @@
...
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie }) => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  // eslint-disable-next-line max-len
+  const defaultPicture = 'https://via.placeholder.com/360x270.png?text=no%20preview';
```
Denchepornyuk, line 21: it should be in the separate file, something like `findMovie.constants.ts`

Denchepornyuk (approved):

PR #834, opened 2022-11-24
Solution for Movies list fetch

sapnachoudhary06:


Bibicutu (approved):
👍

src/App.tsx
```diff
@@ -1,20 +1,33 @@
...
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovie] = useState<Movie[]>([]);
+
+  const addMovie = useCallback((movie: Movie) => {
+    const movieExist = movies.find((m) => (m.imdbId === movie.imdbId));
```
ihor-jpeg, line 11: probably should use .some array method. but this way is also okay

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,67 @@
...
+
+      setIsLoading(false);
+
+      if ('Error' in result) {
+        setIsError(true);
+      } else {
+        setMovie({
+          title: result.Title,
+          description: result.Plot,
+          imgUrl: result.Poster === 'N/A'
```
ihor-jpeg, line 32: better to put N/A in a variable at the top of the files after import, so it's easy to change

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,67 @@
...
+
+  const handleSubmit = useCallback(
+    async (event: React.FormEvent) => {
+      event.preventDefault();
+      setIsLoading(true);
+      const result = await getMovie(query);
+
+      setIsLoading(false);
+
+      if ('Error' in result) {
```
ihor-jpeg, line 26: same as with `N/A`

ihor-jpeg (approved):

ihor-jpeg (approved):

PR #848, opened 2022-12-02
solution

Yaroslav-Radchuk:
[DEMO LINK](https://Yaroslav-Radchuk.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,10 +2,20 @@ import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
-import { Movie } from './types/Movie';
+import { MovieData } from './types/MovieData';
 
-export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+export const App: React.FC = () => {
+  const [movies, setMovie] = useState<(MovieData | null)[]>([]);
+
+  const addMovie = (movie: MovieData | null) => {
+    setMovie(prevState => {
+      if (prevState.find(newMovie => newMovie?.imdbID === movie?.imdbID)) {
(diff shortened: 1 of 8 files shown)


src/api.ts
```diff
@@ -1,9 +1,8 @@
 import { MovieData } from './types/MovieData';
-import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'http://www.omdbapi.com/?i=tt3896198&apikey=fe815c7';
```
AlisaVlasova, line 3: there is some problem with api request
image

AlisaVlasova (changes requested):

src/api.ts
Yaroslav-Radchuk: api key not working? I haven't this problem

AlisaVlasova (changes requested):
I asked another mentor to check demo
We still have same problem with request

src/App.tsx
```diff
@@ -2,10 +2,20 @@ import { useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
-import { Movie } from './types/Movie';
+import { MovieData } from './types/MovieData';
 
-export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+export const App: React.FC = () => {
+  const [movies, setMovie] = useState<(MovieData | null)[]>([]);
```
AlisaVlasova, line 8: image
you need to add Movie type and normalize MovieData to this type

AlisaVlasova (changes requested):

PR #863, opened 2022-12-23
Movies list fetch movies task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,81 @@
-import React from 'react';
+import classNames from 'classnames';
+import React, { useState } from 'react';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieData } from '../../types/MovieData';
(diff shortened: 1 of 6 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,81 @@
...
+        title: response.Title,
+        description: response.Plot,
+        imgUrl: response.Poster !== 'N/A'
+          ? response.Poster
+          : 'https://via.placeholder.com/360x270.png?text=no%20preview',
+        imdbUrl: movieImdbUrl,
+        imdbId: response.imdbID,
+      };
+
+      setIsLoading(false);
```
qoa11a, line 61: You're setting it in finally so no need to repeat it here.

qoa11a (approved):

src/components/FindMovie/FindMovie.tsx
juliiaap: +

PR #865, opened 2022-12-25
solution

MelnychukYana:
https://MelnychukYana.github.io/react_movies-list-fetch-movies/


src/api.ts
```diff
@@ -1,7 +1,7 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=a570192';
```
OlEzhA0, line 4: it's ok use api key like here in test projects but anyway, read this [article](https://create-react-app.dev/docs/adding-custom-environment-variables/) JFYI

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,78 @@
...
+    setTitle(event.target.value);
+    setError(false);
+  };
+
+  const handleGetMovie = (event: React.MouseEvent<HTMLButtonElement,
+  MouseEvent>) => {
+    event.preventDefault();
+    setSearched(true);
+
+    return (title && getMovie(title)
```
OlEzhA0, line 30: why do we need return here?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,78 @@
...
+      .then(data => {
```
OlEzhA0, line 31: Would be better to use `async/await` syntax everywhere

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,78 @@
...
+            title: data.Title,
+            description: data.Plot,
+            imgUrl: data.Poster,
+            imdbId: data.imdbID,
+            imdbUrl: `https://www.imdb.com/title/${data.imdbID}`,
+          });
+
+          setError(false);
+        }
+      })
```
OlEzhA0, line 45: you should `catch` errors as well

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,78 @@
...
+
+    setMovie(null);
+    setTitle('');
+    setError(false);
+  };
+
+  const handleSubmit = (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    event.preventDefault();
```
OlEzhA0, line 71: why do we need to prevent default here?

just add `type="submit"` to your button and `onSubmit={handleSubmit}` to your form (and you can remove onClick event from button)

src/components/MovieCard/MovieCard.tsx
```diff
@@ -6,39 +6,45 @@ type Props = {
...
-            />
-          </figure>
-        </div>
-        <div className="media-content">
-          <p className="title is-8" data-cy="movieTitle">{movie.title}</p>
-        </div>
+const defaultImg = 'https://via.placeholder.com/360x270.png?text=no%20preview';
+
+export const MovieCard: React.FC<Props> = ({ movie }) => {
+  return (
```
OlEzhA0, line 12:
```suggestion
export const MovieCard: React.FC<Props> = ({ movie }) => (
```

OlEzhA0 (approved):
Good job! Please pay attention to my comments and apply changes.
I approve your work, I don't see any critical issues but check `async/await` and `.env`

PR #867, opened 2022-12-26
Solution

OlhaArama:
[DEMO LINK](https://OlhaArama.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,24 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -5,7 +5,24 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie) => {
+    let isMovieInList = false;
+
+    movies.forEach(item => {
```
TarasHoliuk, line 13: it is the perfect case to use `some` method instead of `forEach` (and assign result to const variable)

TarasHoliuk (changes requested):
GJ! Only one thing would be nice to fix

nazarmatsevych (approved):

PR #868, opened 2022-12-29
done

Nazar-Brunarskyi:
[DEMO LINK](https://Nazar-Brunarskyi.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,7 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
 
   return (
     <div className="page">
@@ -14,7 +14,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie onAdd={setMovies} />
       </div>
     </div>
   );
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,88 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { getMovie } from '../../api';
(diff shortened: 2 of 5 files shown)


official-artem (approved):
Seems good. I like it 😎

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,160 @@
...
+    const handleMovieAdding = () => {
+      onAdd((movies) => {
+        if (!movie) {
+          return movies;
+        }
+
+        const movieToAdd = {
+          ...movie,
+        };
+
```
valerii-derkach, line 65: let's move this method to parent component and pass it here as props

and do not forget to display an error when the film is already in the list

valerii-derkach (changes requested):

src/App.tsx
```diff
@@ -5,7 +5,34 @@ import { FindMovie } from './components/FindMovie';
...
+      const movieToAdd = {
+        ...movie,
+      };
+
+      const isMovieAdded = prevMovies.some((film => (
+        film.imdbId === movie.imdbId
+      )));
+
+      if (isMovieAdded) {
+        alert('the movie is already in your list');
```
valerii-derkach, line 25: avoid `alerts`, use some value in state for error and conditional rendering below the query input

src/App.tsx
```diff
@@ -5,7 +5,34 @@ import { FindMovie } from './components/FindMovie';
...
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const onAdd = (movie: Movie) => {
+    setMovies((prevMovies) => {
+      if (!movie) {
+        return prevMovies;
+      }
```
valerii-derkach, line 14: just do not invoke this method in child component, if the is no movie

valerii-derkach (changes requested):

Ihor-Karpyn (approved):

PR #869, opened 2022-12-30
Solution

UmizDemud:
[DEMO LINK](https://UmizDemud.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,11 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const hasMovie = (imdbId: string) => {
+    return !!movies.find(movie => movie.imdbId === imdbId);
+  };
 
   return (
     <div className="page">
@@ -14,7 +18,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie setMovies={setMovies} hasMovie={hasMovie} />
       </div>
     </div>
   );
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,67 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { getMovie } from '../../api';
 import './FindMovie.scss';
+import { MovieCard } from '../MovieCard';
+import { MovieData } from '../../types/MovieData';
+import { movieDataToMovie } from '../../utils/movieDatatoMovie';
+import { ResponseError } from '../../types/ReponseError';
+import { Movie } from '../../types/Movie';
 
-export const FindMovie: React.FC = () => {
+type Props = {
+  setMovies: React.Dispatch<React.SetStateAction<Movie[]>>;
+  hasMovie: (imdbId: string) => boolean;
+};
+
+export const FindMovie: React.FC<Props> = ({ setMovies, hasMovie }) => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<MovieData | null>();
+  const [loading, setLoading] = useState(false);
+  const [isQueryAbsent, setIsQueryAbsent] = useState(true);
+
+  function isMovie(
+    object: MovieData | ResponseError,
+  ): object is MovieData {
+    return (object as MovieData).imdbID !== undefined;
+  }
+
+  const handleAddMovie = () => {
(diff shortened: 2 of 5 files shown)


src/App.tsx
```diff
@@ -5,7 +5,11 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const hasMovie = (imdbId: string) => {
+    return !!movies.find(movie => movie.imdbId === imdbId);
```
naziamoff, line 11: looks like you've implemented `some` method here 😄

naziamoff: and let's call it like the function should be called. I'd make it like `checkIsMovieFound`

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,67 @@
...
+    setMovie(null);
+    setQuery('');
+  };
+
+  const handleQueryChange = (text: string) => {
+    setIsQueryAbsent(true);
+    setQuery(text);
+  };
+
+  const handleMovieRequest = (
```
naziamoff, line 42: looks quite difficult to read. Do you have any ideas on how to improve its readability? Let me know if you got none, I'll help

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +77,59 @@ export const FindMovie: React.FC = () => {
...
 
         <div className="field is-grouped">
           <div className="control">
             <button
               data-cy="searchButton"
               type="submit"
-              className="button is-light"
+              className={classNames(
+                'button',
+                'is-light',
```
naziamoff, line 99: should work this way too I guess
```suggestion
                'button is-light',
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +77,59 @@ export const FindMovie: React.FC = () => {
...
+                { 'is-loading': loading },
+              )}
+              disabled={loading || !query}
+              onClick={(e) => handleMovieRequest(e)}
             >
-              Find a movie
+              {!movie
+                ? 'Find a movie'
+                : 'Search again'}
```
naziamoff, line 107: let's start with positive - that looks more logical

naziamoff (changes requested):
good one, as usual 👍 
But still we can't miss the space for improvements 😄

src/components/FindMovie/FindMovie.tsx
UmizDemud: The parameter isQueryAbsent was starting as true and it was working the opposite way.
It's supposed to mean "is query not found". So I corrected that. 

Nothing else comes to my mind except writing a providePoster helper function to use in setMovie, instead of doing it right there. But I'd like to know if there is something else.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +81,58 @@ export const FindMovie: React.FC = () => {
...
+              onChange={e => handleQueryChange(e.target.value)}
             />
           </div>
 
-          <p className="help is-danger" data-cy="errorMessage">
-            Can&apos;t find a movie with such a title
-          </p>
+          {isQueryAbsent && (
+            <p className="help is-danger" data-cy="errorMessage">
+              Can&apos;t find a movie with such a title
```
meforoma, line 91: actual string is more readable
```suggestion
              {'Can\'t find a movie with such a title'}
              // or
              {`Can't find a movie with such a title`}
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+              ? result.Poster
+              : 'https://via.placeholder.com/360x270.png?text=no%20preview',
+          });
+        } else {
+          setIsQueryAbsent(true);
+        }
+      })
+      .finally(() => {
+        setLoading(false);
+      });
```
meforoma, line 66: the if..else could be declared as a helper 
```
const processMovie = (result) => {
  if (isMovie(result)) {
    setMovie({
      ...result,
      Poster: result.Poster.startsWith('https')
        ? result.Poster
        : 'https://via.placeholder.com/360x270.png?text=no%20preview',
    });
  } else {
    setIsQueryAbsent(true);
  }
}
```

and then
```suggestion
    setLoading(true);
    getMovie(query.trim())
      .then(processMovie)
      //same as `.then(result => callback(result))`
      .finally(() => {
        setLoading(false);
      });
```

meforoma (approved):
essentially just a few suggestions, otherwise the implementation is good 👍🏼

PR #874, opened 2023-01-07
solution

olenitut:
[DEMO LINK](https://olenitut.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,29 +10,31 @@ export const App = () => {
   const [currentMovie, setCurrentMovie] = useState<null | Movie>(null);
   const [isLoading, setIsLoading] = useState(false);
 
-  const findMovieHandler = (userQuery: string) => {
+  // eslint-disable-next-line max-len
+  const defaultPicture = 'https://via.placeholder.com/360x270.png?text=no%20preview';
+
+  const findMovieHandler = async (userQuery: string) => {
     setIsLoading(true);
 
-    getMovie(userQuery).then(data => {
-      if ('Error' in data) {
-        setCurrentMovie(null);
+    const data = await getMovie(userQuery);
 
-        return;
-      }
+    if ('Error' in data) {
+      setCurrentMovie(null);
+      setIsLoading(false);
 
-      // eslint-disable-next-line max-len
-      const defaultPicture = 'https://via.placeholder.com/360x270.png?text=no%20preview';
+      return;
+    }
 
-      const newMovie: Movie = {
-        title: data.Title,
-        description: data.Plot,
-        imgUrl: data.Poster === 'N/A' ? defaultPicture : data.Poster,
-        imdbId: data.imdbID,
-        imdbUrl: `https://www.imdb.com/title/${data.imdbID}`,
-      };
+    const newMovie: Movie = {
+      title: data.Title,
+      description: data.Plot,
+      imgUrl: data.Poster === 'N/A' ? defaultPicture : data.Poster,
+      imdbId: data.imdbID,
+      imdbUrl: `https://www.imdb.com/title/${data.imdbID}`,
+    };
 
-      setCurrentMovie(newMovie);
-    }).finally(() => setIsLoading(false));
+    setCurrentMovie(newMovie);
+    setIsLoading(false);
   };
 
   const addMovieHandler = () => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -46,7 +46,7 @@ export const App = () => {
           return newMovies;
         }
 
-        newMovies.push(currentMovie);
+        return [...prevMovies, currentMovie];
       }
 
       return newMovies;
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,9 +3,55 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -3,9 +3,55 @@ import './App.scss';
...
+      const newMovie: Movie = {
+        title: data.Title,
+        description: data.Plot,
+        imgUrl: data.Poster === 'N/A' ? defaultPicture : data.Poster,
+        imdbId: data.imdbID,
+        imdbUrl: `https://www.imdb.com/title/${data.imdbID}`,
+      };
+
+      setCurrentMovie(newMovie);
+    }).finally(() => setIsLoading(false));
```
OlEzhA0, line 35: try to use `async/await` syntax, it’s more preferable

src/App.tsx
```diff
@@ -3,9 +3,55 @@ import './App.scss';
...
+
+    getMovie(userQuery).then(data => {
+      if ('Error' in data) {
+        setCurrentMovie(null);
+
+        return;
+      }
+
+      // eslint-disable-next-line max-len
+      const defaultPicture = 'https://via.placeholder.com/360x270.png?text=no%20preview';
```
OlEzhA0, line 24: define this constant outside your component

src/App.tsx
```diff
@@ -3,9 +3,55 @@ import './App.scss';
...
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [currentMovie, setCurrentMovie] = useState<null | Movie>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const findMovieHandler = (userQuery: string) => {
+    setIsLoading(true);
+
+    getMovie(userQuery).then(data => {
```
OlEzhA0, line 16: where do you `catching` your error?

src/App.tsx
```diff
@@ -3,9 +3,55 @@ import './App.scss';
...
+  const addMovieHandler = () => {
+    setMovies(prevMovies => {
+      const newMovies = [...prevMovies];
+
+      if (currentMovie) {
+        if (newMovies.find(mov => mov.imdbId === currentMovie.imdbId)) {
+          return newMovies;
+        }
+
+        newMovies.push(currentMovie);
```
OlEzhA0, line 47:
```suggestion
        return [...prevMovies, currentMovie]
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,45 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
 import './FindMovie.scss';
 
-export const FindMovie: React.FC = () => {
+type Props = {
+  onFind: (arg0: string) => void;
```
OlEzhA0, line 8:
```suggestion
  onFind: (value: string) => void;
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,45 @@
...
+  movie: Movie | null;
+  onAdd: () => void;
+  isLoading: boolean
+};
+
+export const FindMovie: React.FC<Props> = ({
```
OlEzhA0, line 14: nb: destruct FC from react

OlEzhA0 (changes requested):
Hey! Please check my comments
ping me in fe_chat if you have any questions

src/App.tsx
olenitut: here     if ('Error' in data) {
      setCurrentMovie(null);
      setIsLoading(false);

      return;
    }

olenitut: check what the api function retruns for errors

src/components/FindMovie/FindMovie.tsx
olenitut: no, i prefere this variant

witflash: Why do you prefer this variant? :)
It's ok, but what is the reason? )

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,45 @@
...
+  onFind, movie, onAdd, isLoading,
+}) => {
+  const [enteredValue, setEnteredValue] = useState('');
+  const [attempteSearch, setAttemptedSearch] = useState(false);
```
witflash, line 18: You have a typo in `attemptedSearch` I think

src/App.tsx
witflash: @olenitut What about this comment?

witflash (changes requested):
Well done!
But don't leave unresolved comments, please.

src/components/FindMovie/FindMovie.tsx
olenitut: I believe that when i use syntax React.FC, it is much more readable and easy to understand where this variable comes from. The same goes with destructuring props. I get a lot of comments from mentors that i need to destructure them but i believe that when i use props.children, the code becomes much more readable for me and others

witflash: You may change your mind when you will work on a huge project :)
But ok )

P. S. Destructuring is a trending best practice. So you should remember this.

witflash (approved):

PR #879, opened 2023-01-09
Develop

frontend112:
frontend112.github.io/react_movies-list-fetch-movies/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,86 @@
-import { useState } from 'react';
+import {
+  useState,
+  ChangeEvent,
+  FC,
+} from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
+import { getMovie } from './api';
 
-export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+export const App: FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const [title, setTitle] = useState('');
+  const [isLoading, setisLoading] = useState(false);
+  const [isTitleCorrect, setIsTitleCorrect] = useState(true);
+  const [movie, setMovie] = useState<Movie | null>(null);
+
+  const handleTitleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    setTitle(event.target.value);
+    setIsTitleCorrect(true);
+  };
+
+  const loadMovie = async () => {
+    try {
+      setIsTitleCorrect(true);
+      setisLoading(true);
+      const loadedMovie = await getMovie(title);
+
+      if (loadedMovie.Response === 'True') {
+        const {
+          Title,
+          Poster,
+          Plot,
+          imdbID,
+        } = loadedMovie;
+
+        const imdbUrl = `https://www.imdb.com/title/${imdbID}`;
+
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,11 +1,90 @@
...
+
+      if (loadedMovie.Response === 'False') {
+        setIsTitleCorrect(false);
+      }
+    } finally {
+      setIsMovieLoading(false);
+    }
+  };
+
+  const handleTtitleSubmit = (event:ChangeEvent<HTMLFormElement>) => {
```
polosanya, line 66:
```suggestion
  const handleTitleSubmit = (event:ChangeEvent<HTMLFormElement>) => {
```

src/App.tsx
```diff
@@ -14,7 +93,15 @@ export const App = () => {
...
-        <FindMovie />
+        <FindMovie
+          handleTtitleSubmit={handleTtitleSubmit}
+          title={title}
+          handleTitleChange={handleTitleChange}
+          isTitleCorrect={isTitleCorrect}
+          movie={movie}
+          isMovieLoading={isMovieLoading}
+          handleAddMovie={handleAddMovie}
+        />
```
polosanya, line 104: Do you really need to pass so many props? Can't you create and use them in FindMovie component instead of App?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +45,71 @@ export const FindMovie: React.FC = () => {
...
+              onChange={handleTitleChange}
             />
           </div>
 
-          <p className="help is-danger" data-cy="errorMessage">
-            Can&apos;t find a movie with such a title
-          </p>
+          {
+            !isTitleCorrect
+            && (
```
polosanya, line 55:
```suggestion
            !isTitleCorrect
              && (
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +45,71 @@ export const FindMovie: React.FC = () => {
...
+        movie
+          && (
+            <div className="container" data-cy="previewContainer">
+              <h2 className="title">Preview</h2>
+              {
+                movie
+                && <MovieCard movie={movie} />
+              }
+            </div>
+          )
```
polosanya, line 111:
```suggestion
        movie
          && (
            <div className="container" data-cy="previewContainer">
              <h2 className="title">Preview</h2>
              
              <MovieCard movie={movie} />
            </div>
          )
```
You've already checked if the movie is present above. Don't duplicate yourself

src/types/MovieData.ts
```diff
@@ -3,4 +3,5 @@ export interface MovieData {
   Title: string;
   Plot: string;
   imdbID: string;
+  Response: 'True';
```
polosanya, line 6: Are you sure that it's the correct type?
Do you really need to hardcore boolean value as string and add a new parameter to MovieData?

polosanya (changes requested):
Almost perfect! Please check the comments to improve your code a bit)

src/types/MovieData.ts
frontend112: @polosanya 
I think that I still need to edit type in MovieData / ResponseError because if I check with hasOwn mathod I am getting errors like in attachment
image

polosanya (approved):
Good job!

PR #888, opened 2023-01-20
solution

batstolya:
[DEMO LINK](https://batstolya.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -23,7 +23,6 @@ export const FindMovie: React.FC<Props> = ({ onMovie }) => {
       const response = await getMovie(query);
 
       if ('Error' in response) {
-        // console.log('errr')
         setError(true);
         setLoading(false);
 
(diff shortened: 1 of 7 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,83 @@
...
+  const handlerAddMovie = () => {
+    onMovie((prev) => {
+      if (!findMovie) {
+        return prev;
+      }
+
+      const hasMovie = (prev.some(movie => movie.imdbId === findMovie?.imdbId));
+
+      if (hasMovie) {
+        alert('you have same');
```
valerii-derkach, line 66: replace alert with error and just show it under the input in markup

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,7 +3,7 @@ import { Movie } from '../../types/Movie';
 import './MovieCard.scss';
 
 type Props = {
-  movie: Movie,
+  movie: Movie | null,
```
valerii-derkach, line 6: do we need null here? just do not render the component if there is no movie

valerii-derkach (changes requested):

PR #892, opened 2023-01-25
solution

Lun4trick:
[DEMO LINK](https://lun4trick.github.io/react_movies-list-fetch-movies/)

I prefer to add debounce always to these kind of imputs, but if i do, the tests get failed.

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,57 +1,138 @@
-import React from 'react';
+import React, { ChangeEvent, useCallback, useState } from 'react';
+import cn from 'classnames';
+import { getMovie } from '../../api';
+import { MovieData } from '../../types/MovieData';
+import { MovieCard } from '../MovieCard';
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,138 @@
...
-          </div>
+type Props = {
+  onAdd: (movie: Movie) => void
+};
+
+export const FindMovie: React.FC<Props> = React.memo(
+  ({ onAdd }) => {
+    const [query, setQuery] = useState('');
+    // eslint-disable-next-line max-len
+    const [searchResponse, setSearchResponse] = useState<MovieData | null>(null);
```
naziamoff, line 17:
```suggestion
    const [
      searchResponse, 
      setSearchResponse,
    ] = useState<MovieData | null>(null);
```

naziamoff (approved):
- nothing happens if I add the same movie to the list twice. Would be nice to show some alert
Everything else looks good 🔥

PR #902, opened 2023-02-06
fetch movies done

Yevhen-Pavlovskyi:
https://yevhen-pavlovskyi.github.io/react_movies-list-fetch-movies/


qoa11a (changes requested):
image
there should be your API key in order for the app to work

anstsot (changes requested):
I've checked the app and it seems not to work properly. Try to check `options` in `fetch` function.
image

lerastarynets (changes requested):
Pls pay attention to previous comments

PR #909, opened 2023-02-15
Add movie form was added

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,79 @@
-import React from 'react';
+import classNames from 'classnames';
+import React, { useState } from 'react';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
 import './FindMovie.scss';
 
-export const FindMovie: React.FC = () => {
+type Props = {
+  onAdd: (newMovie: Movie) => void,
+};
+
+export const FindMovie: React.FC<Props> = ({ onAdd }) => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [isError, setIsError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleFormSubmit = async (event: React.FormEvent) => {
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,79 @@
...
+      setIsError(true);
+    } finally {
+      setIsLoading(false);
+    }
+  };
+
+  const hadleAddButton = () => {
+    if (movie !== null) {
+      onAdd(movie);
+    }
```
Anastasiia-Khudych, line 65: Add button is visible on the page only when a movie isn't null, so this check is redundant.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,79 @@
...
+        const poster = Poster === 'N/A'
+          ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+          : Poster;
+        const newMovie = {
+          title: Title,
+          description: Plot,
+          imgUrl: poster,
+          imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+          imdbId: imdbID,
+        };
```
Anastasiia-Khudych, line 50: Isn't it better to move this code into a separate function, which receives the response and returns the movie?

Anastasiia-Khudych (approved):
Good job :star:

src/components/FindMovie/FindMovie.tsx
denys-cheporniuk: agree, it would be better to move this logic in a separate function

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +86,59 @@ export const FindMovie: React.FC = () => {
               id="movie-title"
               placeholder="Enter a title to search"
               className="input is-dander"
+              value={query}
+              onChange={(event) => {
+                setQuery(event.target.value);
+                setIsError(false);
+              }}
```
denys-cheporniuk, line 93: would be great to move it in a separate function

denys-cheporniuk (approved):

PR #910, opened 2023-02-15
Added the task solution

oleksusov:
[DEMO LINK](https://oleksusov.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,10 +1,17 @@
 import { MovieData } from './types/MovieData';
 import { ResponseError } from './types/ReponseError';
 
-const API_URL = 'https://www.omdbapi.com/?apikey=your-key';
+const API_URL = 'https://www.omdbapi.com/?apikey=e4d19130';
+
+function wait(delay: number): Promise<void> {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
 
 export function getMovie(query: string): Promise<MovieData | ResponseError> {
-  return fetch(`${API_URL}&t=${query}`)
+  return wait(300)
+    .then(() => fetch(`${API_URL}&t=${query}`))
     .then(res => res.json())
     .catch(() => ({
       Response: 'False',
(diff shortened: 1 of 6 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,78 @@
...
+
+      setLoadData(false);
+
+      if ('Error' in movieFromServer) {
+        throw new Error('Can&apos;t find movie such a title');
+      } else {
+        if (movieFromServer.Poster === 'N/A') {
+          movieFromServer.Poster
+            = 'https://via.placeholder.com/360x270.png?text=no%20preview';
+        }
```
witflash, line 40: This case is better to handle inside the `MovieCard` component, not mutate API response.

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,7 +3,7 @@ import { Movie } from '../../types/Movie';
 import './MovieCard.scss';
 
 type Props = {
-  movie: Movie,
+  movie: Movie | undefined,
```
witflash, line 6: Which is the case if you don't have the `movie` prop here?
As I see you check `loadedMovie` and render this component if you have a movie, isn't it?

witflash: There is no reason to render this component if there is no movie.

witflash (changes requested):
Good job!
But check my comments, please.

src/api.ts
```diff
@@ -1,10 +1,17 @@
...
+
+function wait(delay: number): Promise<void> {
+  return new Promise(resolve => {
+    setTimeout(resolve, delay);
+  });
+}
 
 export function getMovie(query: string): Promise<MovieData | ResponseError> {
-  return fetch(`${API_URL}&t=${query}`)
+  return wait(300)
```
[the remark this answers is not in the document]
oleksusov: I didn't need it

qoa11a (approved):

PR #914, opened 2023-02-18
Result

svitjojo:
[DEMO LINK](https://svitjojo.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,19 @@
-import { useState } from 'react';
+import { useCallback, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = useCallback((movie: Movie) => {
+    const hasMovie = movies.some(m => m.imdbId === movie.imdbId);
+
+    if (!hasMovie) {
+      setMovies((currentMovies) => [...currentMovies, movie]);
+    }
+  }, [movies]);
 
   return (
     <div className="page">
@@ -14,7 +22,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie addMovie={addMovie} />
       </div>
     </div>
   );
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,63 @@
-import React from 'react';
+import React, { useCallback, useState } from 'react';
+import cn from 'classnames';
 import './FindMovie.scss';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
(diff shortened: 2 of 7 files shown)


nzrii-kysel (approved):
👍

svitjojo:
Thanks, I appreciate it 🤝

src/App.tsx
```diff
@@ -1,11 +1,19 @@
...
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = useCallback((movie: Movie) => {
+    const hasMovie = movies.some(m => m.imdbId === movie.imdbId);
```
alexandra-protyanova, line 11:
```suggestion
    const hasMovie = movies.some(movie => m.imdbId === movie.imdbId);
```

src/App.tsx
```diff
@@ -1,11 +1,19 @@
...
+
+    if (!hasMovie) {
```
alexandra-protyanova, line 13: it would be nice to show message, why we won't add movie into list, in case of it's already there

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,63 @@
...
+    setIsLoading(true);
+
+    try {
+      const movieFromServer = await getMovie(query);
+
+      if ('Error' in movieFromServer) {
+        setHasError(true);
+        setIsLoading(false);
+
+        return;
```
alexandra-protyanova, line 30: nice case, to put to `catch`

src/types/ResponseError.ts
```diff
@@ -0,0 +1,4 @@
+export interface ResponseError {
+  Response: 'False',
+  Error: string,
+}
```
alexandra-protyanova, line 4: looks like we shouldn't set type, to server's meta data

alexandra-protyanova (changes requested):

src/App.tsx
svitjojo: I can't use movie because it has already been declared.

src/types/ResponseError.ts
svitjojo: This file was already in the folder

src/components/FindMovie/FindMovie.tsx
svitjojo: I can't put it inside catch, because typescript will show an error below when I setMovie and its possible error object.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,69 @@
...
+import { getNormalizeMovie } from '../../utils/getNormalizeMovie';
+import { ErrorMessage } from '../../types/ErrorMessage';
+
+type Props = {
+  addMovie: (movie: Movie) => void,
+  errorMessage: ErrorMessage,
+  changeErrorMessage: (message: ErrorMessage) => void,
+};
+
+export const FindMovie: React.FC<Props> = ({
```
ArturErmolenko, line 16: memo

ArturErmolenko (approved):
GJ 👍

PR #916, opened 2023-02-23

rafmamedov:
[DEMO LINK](https://rafmamedov.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,20 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie) => {
+    const hasMovie = movies
+      .some(addedMovie => addedMovie.imdbId === movie.imdbId);
+
+    if (hasMovie) {
+      return;
+    }
+
+    const updatedMovies = [...movies, movie];
+
+    setMovies(updatedMovies);
+  };
 
   return (
     <div className="page">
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -5,7 +5,20 @@ import { FindMovie } from './components/FindMovie';
...
+    const hasMovie = movies
+      .some(addedMovie => addedMovie.imdbId === movie.imdbId);
+
+    if (hasMovie) {
+      return;
+    }
+
+    const updatedMovies = [...movies, movie];
+
+    setMovies(updatedMovies);
```
SanyaBratashchuk, line 20:
```suggestion
   

    setMovies(prevMovies => [...prevMovies, movie]);
```
It is good practice to pass callback to your setter function if new state value depends on its previous value

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+      imdbUrl,
+      imdbId: response.imdbID,
+    };
+
+    setHasError(false);
+
+    return selectedMovie;
+  };
+
+  const getMovieByName = async (event: { preventDefault: () => void; }) => {
```
SanyaBratashchuk, line 46: Add correct type for events 
```suggestion
  const getMovieByName = async (event: React.MouseEvent<HTMLButtonElement>) => {
```

SanyaBratashchuk (changes requested):

src/App.tsx
rafmamedov: Implemented. Thank you for the review!

anstsot (approved):
Great job, approved!

PR #932, opened 2023-03-16
Solution

dissolveyo:
[DEMO LINK](https://dissolveyo.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -18,17 +18,22 @@ export const FindMovie: React.FC<FindMovieProps> = ({ onMovieAdd }) => {
 
   const handleFindMovie = async () => {
     setIsLoading(true);
-    const movieData = await getMovie(query);
 
-    if ('Error' in movieData) {
+    try {
+      const movieData = await getMovie(query);
+
+      if ('Error' in movieData) {
+        throw Error();
+      } else {
+        const movie = createMovie(movieData);
+
+        setSearchedMovie(movie);
+      }
+    } catch {
       setHasError(true);
-    } else {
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -5,7 +5,21 @@ import { FindMovie } from './components/FindMovie';
...
+    );
+
+    if (isMovieAdded) {
+      return false;
+    }
+
+    setMovies((current) => [...current, newMovie]);
+
+    return true;
+  };
```
IvanFesenko, line 22: Why you return value if you dont use it?
```suggestion
  const handleMovieAdd = (newMovie: Movie) => {
    const isMovieAdded = movies.find(
      (movie) => movie.imdbId === newMovie.imdbId,
    );

    if (!isMovieAdded) {
     setMovies((current) => [...current, newMovie]);
    }
  };
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,62 @@
...
+  const [query, setQuery] = useState('');
+  const [searchedMovie, setSearchedMovie] = useState<Movie | null>(null);
+  const [hasError, setHasError] = useState(false);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleFindMovie = async () => {
+    setIsLoading(true);
+    const movieData = await getMovie(query);
+
+    if ('Error' in movieData) {
```
loralevitska, line 23: Use catch for errors in [`try {...} catch (e) {...}`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch)

loralevitska (approved):

PR #933, opened 2023-03-19
solution without 2 tests to best user experience

webdevw:
I added an option so that the user cannot add two identical names in a row at once, 
so from that two tests crashed


polosanya (changes requested):
User should be able to find the same movie again.
They simply should not be able to add it a second time.
Check how it works in [working example](https://mate-academy.github.io/react_movies-list-fetch-movies/)
movie

src/App.tsx
```diff
@@ -5,7 +5,21 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const pushMovie = (movie: Movie) => {
+    if (movies.length > 0) {
+      const imbdIdMovie = movies
```
polosanya, line 12:
```suggestion
      const isMoviePresent = movies
```
Better specify that it's boolean value to make it clear

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+  const changeMovie = async (inputValue: string) => {
+    setIsLoading(true);
+
+    const newMovie = await getMovie(inputValue)
+      .finally(() => setIsLoading(false));
+
+    if ('Error' in newMovie) {
+      setIsError(true);
+    } else {
+      setIsError(false);
```
polosanya, line 27: You can use then and catch methods instead

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+
+      const {
+        Title, Plot, Poster, imdbID,
+      } = newMovie;
```
polosanya, line 31:
```suggestion
      const {
        Title, 
        Plot, 
        Poster, 
        imdbID,
      } = newMovie;
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+  }, [searchTitle]);
+
+  const sendQuery = (e: React.MouseEvent<HTMLButtonElement>) => {
+    e.preventDefault();
+    const titleEdited = searchTitle.trim().toLowerCase();
+
+    changeMovie(titleEdited);
+  };
+
+  const clearingSideBar = () => {
```
polosanya, line 55:
```suggestion
  const clearSideBar = () => {
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -13,45 +77,64 @@ export const FindMovie: React.FC = () => {
...
-              Add to the list
-            </button>
-          </div>
+          { movie && (
+            <div className="control">
+              <button
+                data-cy="addButton"
+                type="button"
+                className="button is-primary"
+                onClick={() => addToList()}
```
polosanya, line 123:
```suggestion
                onClick={addToList}
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -13,45 +77,64 @@ export const FindMovie: React.FC = () => {
...
       </form>
 
-      <div className="container" data-cy="previewContainer">
-        <h2 className="title">Preview</h2>
-        {/* <MovieCard movie={movie} /> */}
-      </div>
+      { movie && (
+        <div className="container" data-cy="previewContainer">
+          <h2 className="title">Preview</h2>
+          { movie && <MovieCard movie={movie} />}
```
polosanya, line 135:
```suggestion
          <h2 className="title">Preview</h2>
          
          <MovieCard movie={movie} />
```
You've already checked if movie is present above

src/components/MovieCard/MovieCard.tsx
```diff
@@ -6,39 +6,61 @@ type Props = {
...
+  const {
+    imgUrl,
+    title,
+    description,
+    imdbUrl,
+  } = movie;
+
+  const isTitle = (imgUrl !== 'N/A');
+  // why we should use image from another server, but not from this program?
+  // const posterDef = 'images/noPreview.png';
```
polosanya, line 19: I think you can use it

polosanya (changes requested):
Let's improve your code a bit)

src/components/MovieCard/MovieCard.tsx
webdevw: test doesn't work if I use it

src/components/FindMovie/FindMovie.tsx
webdevw: I can't understand how to do it correctly

witflash: Smth like this:
```
try {
  const newMovie = await getMovie(inputValue);
} catch(error) {
  setIsError(true);
} finally {
  setIsLoading(false);
}
```

But if you want to handle `Error` on API response, you can leave this as is

src/components/MovieCard/MovieCard.tsx
```diff
@@ -6,39 +6,61 @@ type Props = {
...
             />
-          </figure>
-        </div>
-        <div className="media-content">
-          <p className="title is-8" data-cy="movieTitle">{movie.title}</p>
-        </div>
+          ) : (
+            <img
+              data-cy="moviePoster"
+              src={posterDef}
```
witflash, line 35: You can use ternary operator just for `src` attribute instead of rendering two different `img`:
```
<img
  data-cy="moviePoster"
  src={isTitle ? imgUrl : posterDef}
  alt="Film logo"
/>
```

src/components/MovieCard/MovieCard.tsx
```diff
@@ -6,39 +6,61 @@ type Props = {
...
-          <figure className="image is-48x48">
+export const MovieCard: React.FC<Props> = ({ movie }) => {
+  const {
+    imgUrl,
+    title,
+    description,
+    imdbUrl,
+  } = movie;
+
+  const isTitle = (imgUrl !== 'N/A');
```
witflash, line 17: It should has `hasImage` name, I think. It this about the image, not the title, am I right?

witflash (approved):
Well done.
I left some uncritical comments. Check it for your future work.

src/components/MovieCard/MovieCard.tsx
webdevw: Ooops =)

PR #947, opened 2023-04-05
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,9 +3,61 @@ import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
+import { getMovie } from './api';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [newMovie, setNewMovie] = useState<Movie | null>(null);
+  const [newTitle, setNewTitle] = useState('');
+  const [isMovieFind, setIsMovieFind] = useState(true);
+  const [isError, setIsError] = useState(false);
+
+  const changeTitle = (event:React.ChangeEvent<HTMLInputElement>) => {
+    setNewTitle(event.target.value);
+    setIsError(false);
+  };
+
+  const findMovie = (event:React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    setIsMovieFind(false);
+
+    getMovie(newTitle)
+      .then(res => {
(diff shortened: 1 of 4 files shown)


polosanya (changes requested):
Pay attention to the tests. Your page shouldn't have "Add to the list" button if movie is not found. 
Same with preview.
Check how it works on [working example](https://mate-academy.github.io/react_movies-list-fetch-movies/)
image
image

Volodymyr-Melnyk-Fron10der (changes requested):
1. do not see your MovieCard in the commit
2. Previous comment was not fixed, there are no commits pushed after review

tetiana-ivanchenko (approved):

tetiana-ivanchenko (changes requested):
If you already search, button should have label "search again"

src/App.tsx
```diff
@@ -3,9 +3,64 @@ import './App.scss';
...
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [newMovie, setNewMovie] = useState<Movie | null>(null);
+  const [newTitle, setNewTitle] = useState('');
+  const [isMovieFind, setIsMovieFind] = useState(true);
+  const [isError, setIsError] = useState(false);
+  const [isFirstSearch, setIsFirstSearch] = useState(true);
+
+  const changeTitle = (event:React.ChangeEvent<HTMLInputElement>) => {
```
witflash, line 16: It should be `changeSearchString` not a title

src/App.tsx
```diff
@@ -3,9 +3,64 @@ import './App.scss';
...
+    setNewTitle(event.target.value);
```
witflash, line 17: And here you should store the user's search string.

src/App.tsx
```diff
@@ -3,9 +3,64 @@ import './App.scss';
...
+
+    getMovie(newTitle)
+      .then(res => {
+        if ('Error' in res) {
+          setNewMovie(null);
+          setIsMovieFind(true);
+          setIsError(true);
+        } else {
+          setNewMovie({
+            title: newTitle,
```
witflash, line 33: You don't need to use a searched text as a title. Use a title from the response as is. And search text you should use only to find the movie by its title or description.

witflash (changes requested):
Good work!
But there is one problem with a movie title. Check my comments, please

polosanya (changes requested):
Almost done! Pay attention to the tests too

PR #950, opened 2023-04-06
Movies List Fetch

Oleksii-Mishchenko:
- [DEMO LINK](https://Oleksii-Mishchenko.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,24 @@
-import { useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
-export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+export const App: React.FC = () => {
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = useCallback((newMovie: Movie) => {
+    const movieIsAdded = movies.some(movie => newMovie.imdbId === movie.imdbId);
+
+    if (movieIsAdded) {
+      return;
+    }
+
+    setMovies(prevMovies => [
+      ...prevMovies,
+      newMovie,
+    ]);
+  }, [movies]);
 
   return (
     <div className="page">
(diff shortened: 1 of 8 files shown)


src/components/MovieCard/MovieCard.tsx
```diff
@@ -6,39 +6,41 @@ type Props = {
...
+        </div>
 
-      <div className="content" data-cy="movieDescription">
-        {movie.description}
-        <br />
-        <a href={movie.imdbUrl} data-cy="movieURL">
-          IMDB
-        </a>
+        <div className="content" data-cy="movieDescription">
+          {movie.description}
```
tetiana-ivanchenko, line 37: use destructuqrization

src/types/Movie.ts
```diff
@@ -1,7 +1,9 @@
-export interface Movie {
-  title: string;
-  description: string;
-  imgUrl: string;
-  imdbUrl: string;
-  imdbId: string;
+export class Movie {
```
tetiana-ivanchenko, line 1: why do you rewrite interface?

tetiana-ivanchenko (changes requested):

src/types/Movie.ts
Oleksii-Mishchenko: Because this way it's easier to create a new Movie
const movie = new Movie(title, description, imgUrl, imdbUrl, imdbId);

src/App.tsx
```diff
@@ -14,7 +27,9 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie
+          addMovie={addMovie}
+        />
```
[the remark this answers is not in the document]
witflash: You can write in a single line if you have only one short property.

src/types/Movie.ts
witflash: There is no sense to use class if you have a static object.
You can create a new movie just this way:
```
const movie = { title, description, imgUrl, imdbUrl, imdbId };
```

So now your `Movie` is not a `type` but this file is in `types/` directory.

PR #952, opened 2023-04-11
Solution

ifuride:
DEMO LINK](https://ifuride.github.io/react_movies-list-fetch-movies/)


tetiana-ivanchenko (changes requested):
in working example, after you find some movie and before you add it, button text should be "search again"

anstsot (approved):
Great job, approved!

PR #959, opened 2023-04-20
Solution by DK

DimaK123:
[DEMO LINK](https://Dimak123.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,84 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
+import { MovieCard } from '../MovieCard';
 import './FindMovie.scss';
 
-export const FindMovie: React.FC = () => {
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+
+type Props = {
+  addMovie: (movie: Movie) => void;
+  moviesList: Movie[]
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie, moviesList }) => {
+  const [titleSearchQuery, setSearchField] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [errorMessage, setErrorMessage] = useState(false);
+  const [loading, setLoading] = useState(false);
+
+  const handleSearchFieldChange
+  = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = event.target;
+
+    setErrorMessage(false);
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,84 @@
...
+type Props = {
+  addMovie: (movie: Movie) => void;
+  moviesList: Movie[]
+};
+
+export const FindMovie: React.FC<Props> = ({ addMovie, moviesList }) => {
+  const [titleSearchQuery, setSearchField] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [errorMessage, setErrorMessage] = useState(false);
+  const [loading, setLoading] = useState(false);
```
tetiana-ivanchenko, line 18: follow convention for the boolean variables names

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +91,55 @@ export const FindMovie: React.FC = () => {
...
           <div className="control">
             <button
               data-cy="searchButton"
               type="submit"
-              className="button is-light"
+              className={classNames(
+                'button is-light',
+                { 'is-loading': loading },
+              )}
+              onClick={() => createMovie()}
```
tetiana-ivanchenko, line 115:
```suggestion
              onClick={createMovie}
```

tetiana-ivanchenko (changes requested):

src/components/FindMovie/FindMovie.tsx
DimaK123: Done

src/components/FindMovie/FindMovie.tsx
DimaK123: It won't work because the function doesn't include the 'event' parameter

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,84 @@
...
+  };
+
+  const createMovie = async (input = titleSearchQuery) => {
+    if (isError) {
+      setError(false);
+    }
+
+    setLoading(true);
+
+    const movieData = await getMovie(input);
```
agniya-i, line 35:
```suggestion
  const createMovie = async () => {
    if (isError) {
      setError(false);
    }

    setLoading(true);

    const movieData = await getMovie(titleSearchQuery);
```

It should work, so you can use `onClick={createMovie}`

agniya-i (approved):

PR #960, opened 2023-04-21
done

jaxxxon78:
[DEMO LINK](https://jaxxxon78.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,7 +1,68 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import classNames from 'classnames';
 
-export const FindMovie: React.FC = () => {
+import { MovieCard } from '../MovieCard';
+import { getMovie } from '../../api';
+import { MovieData } from '../../types/MovieData';
+import { ResponseError } from '../../types/ReponseError';
+import { Movie } from '../../types/Movie';
+
+type Props = {
+  movies: Movie[];
+  query: string;
+  setQuery: React.Dispatch<React.SetStateAction<string>>;
+  setMovies: React.Dispatch<React.SetStateAction<Movie[]>>;
+};
+
+export const FindMovie: React.FC<Props> = ({
+  movies, query, setMovies, setQuery,
+}) => {
+  const [movie, setMovie] = useState<Movie>();
+  const [isError, setIsError] = useState<ResponseError>();
+  const [isLoading, setIsLoading] = useState(false);
+
+  const addMovie = (movieData: MovieData) => (setMovie({
+    title: movieData.Title,
+    description: movieData.Plot,
+    imgUrl: movieData.Poster === 'N/A'
+      ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+      : movieData.Poster,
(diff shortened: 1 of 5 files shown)


OleziO (changes requested):
Deploy the project
image

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,68 @@
...
+
+export const FindMovie: React.FC<Props> = ({
+  movies, query, setMovies, setQuery,
+}) => {
+  const [movie, setMovie] = useState<Movie>();
+  const [isError, setIsError] = useState<ResponseError>();
+  const [isLoading, setIsLoading] = useState(false);
+
+  const addMovie = (movieData: MovieData) => (setMovie({
+    title: movieData.Title,
```
Anastasiia-Svintsova, line 26: use destructuring for the movieData object

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,68 @@
...
+      ({ imdbId }) => imdbId === movieInfo.imdbId,
+    );
+
+    if (isAlreadyAdded) {
+      return;
+    }
+
+    setMovies([...movies, movieInfo]);
+    setQuery('');
+    setMovie(undefined);
```
Anastasiia-Svintsova, line 63: It's better to use null instead of undefined.
```suggestion
    setMovie(null);
```

Null is a value that can be assigned to a variable, whereas undefined is a default value assigned to a variable that has not been assigned a value yet. This means that if you want to indicate that a variable has no value, it's better to assign it the value of null explicitly rather than relying on the default undefined value.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -30,27 +95,40 @@ export const FindMovie: React.FC = () => {
             <button
               data-cy="searchButton"
               type="submit"
-              className="button is-light"
+              className={classNames('button is-light',
+                { 'is-loading': isLoading })}
+              onClick={(event) => {
+                event.preventDefault();
+                getMovieData(query);
+              }}
```
Anastasiia-Svintsova, line 103: Move it to a function

Anastasiia-Svintsova (changes requested):
- The movie should just disappear from the preview if it already exists in the list. Check how it works in the [working example](https://mate-academy.github.io/react_movies-list-fetch-movies/) and compare it with yours.

https://user-images.githubusercontent.com/99403198/233780131-f41dcde2-875d-4f64-86f2-bde169144b18.mov

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,11 +88,15 @@ export const FindMovie: React.FC = () => {
               id="movie-title"
               placeholder="Enter a title to search"
               className="input is-dander"
+              value={query}
+              onChange={(event) => {
```
TarasHoliuk, line 92: move it to a separate function
Also, there you should reset `isError`

TarasHoliuk (changes requested):
GJ :+1: 
But one thing still should be fixed (check comment)

tetiana-ivanchenko (changes requested):
image
should be "Search again" if I already searched for some movie

IvanFesenko (approved):

PR #971, opened 2023-05-03
The search-movie app has been made

bojkovladislav:


src/App.tsx
```diff
@@ -1,20 +1,75 @@
...
+  const [query, setQuery] = useState<string>('');
+  const [isSubmitted, setIsSubmitted] = useState<boolean>(false);
+  const [showError, setShowError] = useState<boolean>(false);
+
+  const isMovieData = (data: MovieData | ResponseError): data is MovieData => {
+    return (data as MovieData).Title !== undefined;
+  };
+
+  useEffect(() => {
+    if (isSubmitted && query.length) {
```
Anastasiia-Svintsova, line 24: It's not a good way to handle submitting functions by useEffect. Just create a function and call it on the submit button click.

src/App.tsx
```diff
@@ -1,20 +1,75 @@
...
+      getMovie(query)
+        .then((movieFromServer) => {
+          if (isMovieData(movieFromServer)) {
+            setMovie({
+              title: movieFromServer.Title,
```
Anastasiia-Svintsova, line 29: Use destructuring for the movieFromServer object

src/App.tsx
```diff
@@ -1,20 +1,75 @@
...
+          setIsSubmitted(false);
+        });
+    }
+  }, [isSubmitted]);
 
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={movies} />
+        {movies.length > 0 && (
```
Anastasiia-Svintsova, line 55: You can use logical operators instead 
```suggestion
        {!!movies.length && (
```

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,58 @@
...
+  setShowError: (showError: boolean) => void;
+}
+
+export const FindMovie: React.FC<Props> = ({
+  query, setQuery, movie, setMovies, movies, setIsSubmitted, isSubmitted,
+  setMovie, showError, setShowError,
+
+}) => {
+  useEffect(() => {
+    setShowError(false);
```
Anastasiia-Svintsova, line 27: Add a condition here. You need to set the error to false only if there is an error. Don't call the setShowError method on each query change

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,58 @@
...
+
+      if (movies.length
+        && movies.some(({ imdbId }) => movie.imdbId === imdbId)) {
+        setMovies(movies);
+      }
+    }
+
+    setQuery('');
+    setMovie(null);
+  };
```
Anastasiia-Svintsova, line 48: You can move this function to App.tsx to not pass such a lot of props to this component. But it's up to you.

Anastasiia-Svintsova (changes requested):

agniya-i (approved):

PR #972, opened 2023-05-03
solution

toljan1:
[DEMO LINK](https://toljan1.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,45 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
+import { getMovie } from './api';
+import { ResponseError } from './types/ReponseError';
+import { MovieData } from './types/MovieData';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [query, setQuery] = useState('');
+  const [showError, setShowError] = useState(false);
+  const [someMovie, setMovie] = useState<Movie>();
+  const [isRequested, setIsRequested] = useState(false);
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -14,7 +48,19 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie
+          setQuery={setQuery}
+          movieFound={someMovie}
+          setMovies={setMovies}
+          setMovie={setMovie}
+          query={query}
```
mykhalenych, line 56: Pass the 'movies' state to 'FindMovie' instead of the 'setMovies' function to handle duplicate prevention within 'App' component.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +52,68 @@ export const FindMovie: React.FC = () => {
...
-            </button>
-          </div>
+          {movieFound && (
+            <div className="control">
+              <button
+                data-cy="addButton"
+                type="button"
+                className="button is-primary"
+                onClick={() => {
+                  setMovies(
```
mykhalenych, line 95: To prevent adding a movie to the list twice, check if the movie already exists in the list before updating the state.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -17,41 +52,68 @@ export const FindMovie: React.FC = () => {
...
           <div className="control">
             <button
               data-cy="searchButton"
               type="submit"
-              className="button is-light"
+              className={classNames(
+                'button is-light',
+                { 'is-loading': isNowRequested },
+              )}
+              onClick={(event) => {
```
mykhalenych, line 76: Consider using 'onSubmit' on the form element instead of using 'onClick' on the submit button.

mykhalenych (changes requested):
Screenshot 2023-05-03 at 18 29 05
after success submit I got an error

PR #1006, opened 2023-05-11
solution

ladrom:
[demo](https://ladrom.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,7 +5,17 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (newMovie: Movie) => {
+    const repeat = movies.some(movie => movie.imdbId === newMovie.imdbId);
+
+    if (repeat) {
+      return;
+    }
+
+    setMovies([...movies, newMovie]);
+  };
 
   return (
     <div className="page">
@@ -14,7 +24,7 @@ export const App = () => {
       </div>
 
       <div className="sidebar">
-        <FindMovie />
+        <FindMovie addMovie={addMovie} />
       </div>
     </div>
   );
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,69 @@
-import React from 'react';
+import React, { useState } from 'react';
+import classNames from 'classnames';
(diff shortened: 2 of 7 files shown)


src/App.tsx
```diff
@@ -5,7 +5,17 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (newMovie: Movie) => {
+    const repeat = movies.some(movie => movie.imdbId === newMovie.imdbId);
```
polosanya, line 11:
```suggestion
    const isMoviePresent = movies.some(movie => movie.imdbId === newMovie.imdbId);
```
It would be great to give more meaningful name, so it's clear that it is a boolean value
https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,42 +3,44 @@ import { Movie } from '../../types/Movie';
 import './MovieCard.scss';
 
 type Props = {
-  movie: Movie,
+  movie: Movie | null,
```
polosanya, line 6: Are you sure that movie can be null? 
You render MovieCard only if movie is present, so movie won't be null anyway
image

src/components/MovieCard/MovieCard.tsx
```diff
@@ -3,42 +3,44 @@ import { Movie } from '../../types/Movie';
...
-            <img
-              src="images/imdb-logo.jpeg"
-              alt="imdb"
-            />
-          </figure>
-        </div>
-        <div className="media-content">
-          <p className="title is-8" data-cy="movieTitle">{movie.title}</p>
-        </div>
+  movie && (
```
polosanya, line 10: This check is actually redundant. Make the movie the necessary prop

polosanya (changes requested):
Almost perfect! A few small but important fixes left

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,69 @@
...
+    if (!movie) {
+      return;
+    }
+
+    addMovie(movie);
+    setMovie(null);
+    setQuery('');
+  };
+
+  const emptyInput = query.trim().length === 0;
```
Anastasiia-Svintsova, line 59:
```suggestion
  const emptyInput = !query.trim().length;
```

Anastasiia-Svintsova (approved):

PR #1020, opened 2023-05-27
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,72 @@
-import React from 'react';
+import React, { useState } from 'react';
 import './FindMovie.scss';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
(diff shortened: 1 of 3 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,72 @@
...
+  const [errorMessage, setErrorMessage] = useState(false);
+
+  const loadData = () => {
+    setIsLoading(true);
+
+    return getMovie(query)
+      .then(data => {
+        if ('Error' in data) {
+          setErrorMessage(true);
+        }
```
OleziO, line 22: You do not need this cause you have a catch block

OleziO (changes requested):
image
Fix input styles (delete border when data is correct)

src/components/FindMovie/FindMovie.tsx
yuran007: 👌, I understood.

mykhalenych (approved):

PR #1021, opened 2023-05-30
Solution

8Melya8:
[DEMO LINK](https://8melya8.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,82 @@
-import React from 'react';
+import React, {
+  FormEvent, useCallback, useMemo, useState,
+} from 'react';
+import classNames from 'classnames';
 import './FindMovie.scss';
+import { Movie } from '../../types/Movie';
+import { ResponseError } from '../../types/ReponseError';
+import { getMovie } from '../../api';
+import { MovieCard } from '../MovieCard';
 
-export const FindMovie: React.FC = () => {
+interface Props {
+  addToList: (movie: Movie) => void,
+}
+
+export const FindMovie: React.FC<Props> = ({ addToList }) => {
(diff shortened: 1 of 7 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,82 @@
...
+  }, [query]);
+
+  const handleQueryChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      setQuery(event.target.value);
+      setError(null);
+    }, [],
+  );
+
+  const buttonDisabled = useMemo(() => query === '', [query]);
```
etojeDenys, line 67: redundant to cover all vars or functions in useMemo or useCallback

use these hooks only when necessary

etojeDenys: 1. Expensive computations: If you have a function that performs heavy calculations or data processing, you can use useMemo to memoize the result and avoid recomputing it on every render.
2. Rendering optimization: If a component depends on a value that doesn't change often, you can use useMemo to memoize the derived value and prevent unnecessary re-renders.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,78 @@
...
+      setLoading(false);
+    }
+  }, [query]);
+
+  const handleQueryChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setQuery(event.target.value);
+    setError(null);
+  };
+
+  const buttonDisabled = () => query === '';
```
Viktor-Kost, line 63: probably, no needs to create a function to check this query. You can use `query === ''` directly for your button

Viktor-Kost (approved):

PR #1041, opened 2023-06-28
Solution

bohdan-mykhailenko:
[DEMO LINK](https://bohdan-mykhailenko.github.io/react_movies-list-fetch-movies/)


roman-mirzoian (changes requested):
The task is done well and the application itself works very well, but there is a small comment on the code and the output from the task: now you can add movies that are already in the list of added movies several times.
image

naziamoff (approved):
🔥

PR #1042, opened 2023-06-29
add task solution

KatyaZaremska:
[DEMO LINK](https://KatyaZaremska.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,72 @@
-import React from 'react';
+import React, {
+  FormEvent, useCallback, useMemo, useState,
+} from 'react';
 import './FindMovie.scss';
+import classNames from 'classnames';
+import { ResponseError } from '../../types/ReponseError';
+import { getMovie } from '../../api';
+import { Movie } from '../../types/Movie';
+import { MovieCard } from '../MovieCard';
 
-export const FindMovie: React.FC = () => {
+interface Props {
+  addToList: (movie: Movie) => void;
+}
+
+export const FindMovie: React.FC<Props> = ({ addToList }) => {
+  const [query, setQuery] = useState('');
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [error, setError] = useState<ResponseError | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const handleSubmit = useCallback((event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+  }, []);
+
+  const handleSearchSubmit = useCallback(async () => {
+    try {
+      setIsLoading(true);
+
+      const movieData = await getMovie(query);
+
+      if ('Error' in movieData) {
+        setError(movieData);
+        setMovie(null);
+      } else {
+        const {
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,72 @@
...
+        setMovie({
+          title: Title,
+          description: Plot,
+          imgUrl: Poster,
+          imdbId: imdbID,
+          imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+        });
+
+        setError(null);
+      }
```
bohdan-mykhailenko, line 48: It's not the best practice to use if else structure, you definitely can simplify that to make the code more clear and readable.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +78,62 @@ export const FindMovie: React.FC = () => {
...
-          <div className="control">
-            <button
-              data-cy="addButton"
-              type="button"
-              className="button is-primary"
-            >
-              Add to the list
-            </button>
-          </div>
+          {!!movie && (
```
bohdan-mykhailenko, line 109: Why do you use '!' operator twice, it's not clear how it must work.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +78,62 @@ export const FindMovie: React.FC = () => {
...
+            </div>
+          )}
         </div>
       </form>
 
-      <div className="container" data-cy="previewContainer">
-        <h2 className="title">Preview</h2>
-        {/* <MovieCard movie={movie} /> */}
-      </div>
+      {!!movie && (
```
bohdan-mykhailenko, line 130: The same with '!' is here

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +78,62 @@ export const FindMovie: React.FC = () => {
...
+                data-cy="addButton"
+                type="button"
+                className="button is-primary"
+                onClick={() => {
+                  if (movie !== null) {
+                    addToList(movie);
+                    setQuery('');
+                    setMovie(null);
+                  }
+                }}
```
bohdan-mykhailenko, line 121: You can create a separate function to handle that event and make the code more clear

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,72 @@
...
+          description: Plot,
+          imgUrl: Poster,
+          imdbId: imdbID,
+          imdbUrl: `https://www.imdb.com/title/${imdbID}`,
+        });
+
+        setError(null);
+      }
+    } catch {
+      setError({ Response: 'False', Error: 'unexpected error' });
```
bohdan-mykhailenko, line 50: You use here the same code as in your API file to catch the error 
 
 .catch(() => ({
      Response: 'False',
      Error: 'unexpected error',
    }));

bohdan-mykhailenko (changes requested):
Good job 👍 
But the DEMO link doesn't work
And here are some points to improve code style and readability

PR #1046, opened 2023-07-02
Solution

julshvets:
[DEMO LINK](https://julshvets.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -1,10 +1,84 @@
-import React from 'react';
+import React, { useMemo, useState } from 'react';
 import './FindMovie.scss';
+import classNames from 'classnames';
+import { getMovie } from '../../api';
+import { MovieData } from '../../types/MovieData';
+import { MovieCard } from '../MovieCard';
+import { Movie } from '../../types/Movie';
+import { normalizeMovieData } from '../../helpers';
 
-export const FindMovie: React.FC = () => {
+interface Props {
+  setMovies: React.Dispatch<React.SetStateAction<Movie[]>>;
+  movies: Movie[];
+}
+
+export const FindMovie: React.FC<Props> = ({
+  setMovies,
+  movies,
+}) => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const [movieData, setMovieData] = useState<MovieData | null>(null);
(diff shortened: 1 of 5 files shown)


silchyn (changes requested):
- Should the input be invalid initially?
- Show some hints if a user tries to add a movie to the list twice! (currently, submission clears the form)
- You can consider a query consisting of spaces empty

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,98 @@
...
+
+  const handleChangeSearchInput = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setSearchQuery(event.target.value);
+    setIsError(false);
+    setErrorMessage(null);
+  };
+
+  const handleSubmitButton = (
```
naziamoff, line 33: we never handle buttons

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,98 @@
...
+  setMovies,
+  movies,
+}) => {
+  const [searchQuery, setSearchQuery] = useState('');
+  const [movieData, setMovieData] = useState<MovieData | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const [isError, setIsError] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+
+  const handleChangeSearchInput = (
```
naziamoff, line 25: usually we wrap callbacks into the `useCallback` hooks. This way you have it in your code now - we re-create this callback on each render, even if we don't have any reason for that

src/helpers.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  const baseImdbUrl = 'https://www.imdb.com/title/';
+  const defaultImgUrl = (
+    'https://via.placeholder.com/360x270.png?text=no%20preview'
+  );
+
+  return ({
+    title: Title,
+    description: Plot,
+    imgUrl: Poster !== 'N/A' ? Poster : defaultImgUrl,
```
naziamoff, line 19: should be moved to a variable. You should not handle data inside the return in this case

naziamoff (approved):
GJ! A few suggestions

PR #1048, opened 2023-07-03
add functionality to fetch-movies

cephy77:
[DEMO LINK](https://cephy77.github.io/react_movies-list-fetch-movies/)


silchyn (changes requested):
Show some hints if a user tries to add a movie to the list twice! (currently, submission clears the form)

PR #1056, opened 2023-07-11
1st version

Grechana:
https://Grechana.github.io/react_movies-list-fetch-movies

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -9,9 +9,10 @@ export const App = () => {
 
   const handleAddMovie = (movie: Movie) => {
     setMovies(prevState => {
-      const movieExist = prevState.find(state => state.imdbId === movie.imdbId);
+      const currentMovie = prevState
+        .find(state => state.imdbId === movie.imdbId);
 
-      return movieExist ? prevState : [
+      return currentMovie ? prevState : [
         ...prevState,
         movie,
       ];
--- a/src/api.ts
+++ b/src/api.ts
@@ -26,7 +26,7 @@ export const newMovie = (movieData: MovieData): Movie => {
     description: Plot,
     imgUrl: Poster === 'N/A'
       ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
-      : movieData.Poster,
+      : Poster,
     imdbUrl: `https://www.imdb.com/title/${imdbID}`,
     imdbId: imdbID,
   };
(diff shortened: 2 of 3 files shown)


OleziO (changes requested):
Remove red border
image

src/api.ts
```diff
@@ -11,3 +12,22 @@ export function getMovie(query: string): Promise<MovieData | ResponseError> {
...
+    Poster,
+    imdbID,
+  } = movieData;
+
+  return {
+    title: Title,
+    description: Plot,
+    imgUrl: Poster === 'N/A'
+      ? 'https://via.placeholder.com/360x270.png?text=no%20preview'
+      : movieData.Poster,
```
lerastarynets, line 29: You're already destructuring 'Poster' from 'movieData', there's no need to access it again through 'movieData.Poster', simply use 'Poster'.

src/App.tsx
```diff
@@ -5,7 +5,18 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const handleAddMovie = (movie: Movie) => {
+    setMovies(prevState => {
+      const movieExist = prevState.find(state => state.imdbId === movie.imdbId);
```
lerastarynets, line 12: Prefer using meaningful variable name instead of 'state' in the find method. It would be more appropriate to use 'existingMovie' or 'currentMovie'.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +56,59 @@ export const FindMovie: React.FC = () => {
...
-            </button>
-          </div>
+          {movie && (
+            <div className="control">
+              <button
+                data-cy="addButton"
+                type="button"
+                className="button is-primary"
+                onClick={() => {
+                  handleClear();
```
lerastarynets, line 95: You are clearing the form before adding the movie to the list. When handleAddMovie is asynchronous this could lead to race conditions where the movie is not added. It would be safer to clear the form after the movie has been added successfully.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +56,59 @@ export const FindMovie: React.FC = () => {
...
+                  handleAddMovie(movie);
+                }}
```
lerastarynets, line 97: Pls move the function inside `onClick` handler to a separate const and use as a callback here

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit more

polosanya (approved):

PR #1062, opened 2023-07-16
add solution

allagutsul27:
[DEMO LINK](https://allagutsul27.github.io/react_movies-list-fetch-movies/)


VitaliyBondarenko1982 (changes requested):
Try to fix tests.
Screenshot from 2023-07-16 16-49-20
When movie exist in the list - preview and input should be cleared

VitaliyBondarenko1982 (approved):
Well done

PR #1066, opened 2023-07-18
Solution

Ssviatt:
[DEMO LINK](https://<Ssviatt>.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,11 +1,50 @@
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import './App.scss';
 import { MoviesList } from './components/MoviesList';
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
+import { MovieData } from './types/MovieData';
+
+import { getMovie } from './api';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [formInput, setFormInput] = useState('');
+  const [newMovies, setNewMovies] = useState<MovieData | null>(null);
+
+  const changeInput = (value: string) => {
+    setFormInput(value);
+  };
+
+  const moviesQuery = (e: React.MouseEvent<HTMLButtonElement>) => {
+    getMovie(formInput).then(res => {
+      if (!Object.hasOwnProperty.call(res, 'Error')) {
+        setNewMovies(res as MovieData);
+      } else {
+        setNewMovies(null);
+      }
+    }).finally(() => {
+      const t = e.target as HTMLElement;
+
+      t.classList.remove('is-loading');
+    });
+  };
+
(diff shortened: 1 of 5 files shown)


etojeDenys (changes requested):
we don't need to see the error in this case

https://github.com/mate-academy/react_movies-list-fetch-movies/assets/70220667/8bd40fa4-974c-4d46-ac69-ab289d22378a

src/App.tsx
```diff
@@ -1,11 +1,48 @@
...
 import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
+import { MovieData } from './types/MovieData';
+
+import { getMovie } from './api';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+  const [formInput, setFormInput] = useState('');
```
DarkMistyRoom, line 12: do we really need to store formInput in the App, or it can be moved to FindMovie?

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,41 +57,60 @@ export const FindMovie: React.FC = () => {
...
               id="movie-title"
               placeholder="Enter a title to search"
-              className="input is-danger"
+              className={classNames('input', {
+                'is-danger': !movie && touched && !loading,
+              })}
+              value={formInput}
+              onChange={(e) => {
+                onChangeInput(e.target.value);
+                setTouched(false);
```
OleziO, line 66: Create a separated function and path it here.

src/App.tsx
DarkMistyRoom: I ask because I want you to consider this point

src/App.tsx
```diff
@@ -1,11 +1,48 @@
...
+  const [newMovies, setNewMovies] = useState<MovieData | null>(null);
```
DarkMistyRoom, line 13:
```suggestion
  const [newMovie, setNewMovie] = useState<MovieData | null>(null);
```

DarkMistyRoom: looks like we can have only one newMovie

DarkMistyRoom: consider moving this state to the `FindMovie` too

src/App.tsx
```diff
@@ -1,11 +1,48 @@
...
+  const [loading, setLoading] = useState(false);
+
+  const changeInput = (value: string) => {
+    setFormInput(value);
+  };
```
DarkMistyRoom, line 18: please check if we can move this or any other methods to the component too. No need to lift a state where it isn't necessary

DarkMistyRoom (changes requested):
Good job! Let's improve it a bit:
- Don't add the container if a movie isn't found
image
image

- And fix these two tests too
image

SanyaBratashchuk (approved):

PR #1079, opened 2023-07-26
Solution

outofs:
https://outofs.github.io/react_movies-list-fetch-movies/


anstsot (approved):
Good work, approving:sparkles:

PR #1087, opened 2023-07-30

vintwp:
https://vintwp.github.io/react_movies-list-fetch-movies/


src/App.tsx
```diff
@@ -5,7 +5,22 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (newMovie: Movie) => {
+    if (movies) {
```
volodymyr-soltys97, line 11: The movies variable is an array and all arrays are truthy in JavaScript, so this condition is not necessary. Arrays, even if they are empty, are always truthy.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,165 @@
...
+export const FindMovie: React.FC<Props> = ({ onAdd }) => {
+  const [movie, setMovie] = useState<Movie | null>(null);
+  const [searchFieldValue, setSearchFieldValue] = useState<string>('');
+  const [searchName, setSearchName] = useState<string>('');
+  const [isSearching, setIsSearching] = useState<boolean>(false);
+  const [isError, setIsError] = useState<boolean>(false);
+  const searchField = useRef<HTMLInputElement | null>(null);
+
+  const handlerChangeSearchField = (event: HTMLInputElement) => {
+    if (event.value.length > 0) {
```
volodymyr-soltys97, line 22: The argument type of the 'handlerChangeSearchField' function should be 'React.ChangeEvent<HTMLInputElement>' instead of 'HTMLInputElement'.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,165 @@
...
             <input
+              ref={searchField}
+              value={searchFieldValue}
               data-cy="titleField"
               type="text"
               id="movie-title"
               placeholder="Enter a title to search"
-              className="input is-danger"
+              className={cn('input', { 'is-danger': isError })}
+              onChange={e => handlerChangeSearchField(e.target)}
```
volodymyr-soltys97, line 109: The event in the 'onChange' handler should be typed as 'React.ChangeEvent<HTMLInputElement>'.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,165 @@
...
+    setSearchFieldValue(event.value);
+  };
+
+  const handlerSubmitSearch = (e: React.FormEvent) => {
+    e.preventDefault();
+    setSearchName(searchFieldValue);
+  };
+
+  const handlerReceivedData = (data: MovieData) => {
+    const tempPoster
```
volodymyr-soltys97, line 35: Try to avoid hardcoding URLs in your components. Consider moving the placeholder URL to a constants file.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

src/App.tsx
vintwp: Hello. Thanks for your attention about my mistake. Would you suggest - How to avoid this warning in typescript correctly.
At now, i've fixed codition to 
`if (movies.length > 0) {}`

volodymyr-soltys97 (approved):
Well done 🔥

PR #1088, opened 2023-07-31
solution

vlkzmn:
[DEMO LINK](https://vlkzmn.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -5,16 +5,29 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie): void => {
+    const isInList = movies.some(item => item.imdbId === movie.imdbId);
+
+    if (!isInList) {
+      setMovies([
+        ...movies,
+        movie,
+      ]);
+    }
+  };
 
   return (
     <div className="page">
       <div className="page-content">
-        <MoviesList movies={movies} />
+        {!!movies.length && (
+          <MoviesList movies={movies} />
+        )}
       </div>
 
       <div className="sidebar">
-        <FindMovie />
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -5,16 +5,29 @@ import { FindMovie } from './components/FindMovie';
 import { Movie } from './types/Movie';
 
 export const App = () => {
-  const [movies] = useState<Movie[]>([]);
+  const [movies, setMovies] = useState<Movie[]>([]);
+
+  const addMovie = (movie: Movie): void => {
+    const isInList = movies.some(item => item.imdbId === movie.imdbId);
```
lerastarynets, line 11: It's better not to use 'isInList' as a variable name. The name should reflect its purpose clearly. You can use something like 'isMovieInList' which provides more context.

src/components/FindMovie/FindMovie.tsx
```diff
@@ -16,42 +62,56 @@ export const FindMovie: React.FC = () => {
...
+              className="input"
+              value={query}
+              onChange={(event) => setQuery(event.target.value)}
             />
           </div>
 
-          <p className="help is-danger" data-cy="errorMessage">
-            Can&apos;t find a movie with such a title
-          </p>
+          {movie === false && query === prevQuery.current && (
```
lerastarynets, line 71:
```suggestion
          {!movie && query === prevQuery.current && (
```

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

For some reason error is not cleared after user enter new search query

https://github.com/mate-academy/react_movies-list-fetch-movies/assets/77551820/a7323c1a-1375-422b-9cd1-d50dacd3b400

src/components/FindMovie/FindMovie.tsx
vlkzmn: I need here **movie === false**, because movie can be null
**const [movie, setMovie] = useState<Movie | null | false>(null);**

lerastarynets (approved):
You're doing a fantastic job, keep going!

PR #1097, opened 2023-08-06
The task is completed

AndruhaMan:
[DEMO LINK](https://AndruhaMan.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,15 +15,8 @@ export const App = () => {
 
       <div className="sidebar">
         <FindMovie
-          onMovieAdd={(movie) => {
-            const haveThisMovieInList = movies.some(movieInList => (
-              movieInList.imdbId === movie.imdbId
-            ));
-
-            if (!haveThisMovieInList) {
-              setMovies([...movies, movie]);
-            }
-          }}
+          movies={movies}
+          onMovieAdd={(movie) => setMovies([...movies, movie])}
         />
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,73 @@
...
+  const [foundMovie, setFoundMovie]
+    = useState<Movie | null>(null);
+
+  const handleSubmit = (event: React.FormEvent) => {
+    event.preventDefault();
+    setIsLoading(true);
+
+    getMovie(query)
+      .then((data: MovieData | ResponseError) => {
+        if ('Error' in data) {
```
volodymyr-soltys97, line 30: The error handling in the 'getMovie' promise is not ideal. If the promise rejects, the 'Error' property won't necessarily be in the data. Consider using a '.catch()' block to handle errors more effectively.

src/App.tsx
```diff
@@ -14,7 +14,17 @@ export const App = () => {
...
+        <FindMovie
+          onMovieAdd={(movie) => {
+            const haveThisMovieInList = movies.some(movieInList => (
+              movieInList.imdbId === movie.imdbId
+            ));
+
+            if (!haveThisMovieInList) {
+              setMovies([...movies, movie]);
+            }
+          }}
```
volodymyr-soltys97, line 26: It is better to avoid such logic in props, you can write a helper function for this and pass it

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

polosanya (approved):
Good job!

PR #1101, opened 2023-08-10
Movies List

dudar-y:
[DEMO LINK](https://dudar-y.github.io/react_movies-list-fetch-movies/)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,85 @@
...
+
+  useEffect(() => {
+    if (data !== null && 'Title' in data) {
+      const newMovie: Movie = {
+        title: data.Title,
+        description: data.Plot,
+        imgUrl: (!data.Poster || data.Poster === 'N/A')
+          ? DEFAULT_POSTER
+          : data.Poster,
+        imdbUrl: `https://www.imdb.com/title/${data.imdbID}`,
```
VitaliyBodnar, line 73: it's better to move 'https://www.imdb.com/title/' to the constant, because it's a static data

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,85 @@
...
+  const [isLoading, setIsLoading] = useState(false);
+  const [data, setData] = useState<MovieData | null | ResponseError>(null);
+  const [foundMovie, setFoundMovie] = useState<Movie | null>(null);
+
+  const handleQueryChange = (
+    event: React.ChangeEvent<HTMLInputElement>,
+  ) => {
+    setQuery(event.target.value);
+    if (isError) {
+      setIsError(false);
```
VitaliyBodnar, line 29: maybe `setIsError(true)` ?

VitaliyBodnar (changes requested):
Almost done!

src/components/FindMovie/FindMovie.tsx
dudar-y: Why? During input we check if there is an error (true) and if there is, we reset the error value (false) to remove the error message. I think the logic is correct, or not?..

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,87 @@
...
+    }
```
TarasHoliuk, line 32: yeah, the logic is correct, but it might look a bit confusing after the first look)
you can just set it false w/o checking: [article](https://pgarciacamou.medium.com/react-doesnt-always-trigger-a-re-render-on-setstate-4644212560a)

TarasHoliuk (changes requested):
GJ!
Let's fix just one thing :)

etojeDenys (approved):
lgtm

PR #1121, opened 2023-09-10
search and add movies from the database

w315:
[DEMO LINK](https://w315.github.io/react_movies-list-fetch-movies/)

Everything seems to be working fine, but for some reason three tests do not pass, I do not understand how to fix them, and when npm run deploy writes an error (I wrote about it in fe_chat)

Error:
FindMovie component:
1) should show a spinner while waiting for the search results
2) should hide a spinner after search response

Add button:
3) should allow to add one more movie to the list


volodymyr-soltys97 (changes requested):
Problem with demo link, you need to make a deploy
image

PR #1124, opened 2023-09-11
add list

w315:
[DEMO LINK](https://w315.github.io/react_movies-list-fetch-movies/)

Everything seems to be working fine, but for some reason three tests do not pass, I do not understand how to fix them, and when npm run deploy writes an error (I wrote about it in fe_chat)

Error:
FindMovie component:
- should show a spinner while waiting for the search results
- should hide a spinner after search response

Add button:
- should allow to add one more movie to the list


volodymyr-soltys97 (changes requested):
Need to fix it
image

anastasiiavorobiova (approved):
Good work!

PR #1125, opened 2023-09-13
solution

BipBobBoop:
[DEMO LINK](https://BipBobBoop.github.io/react_movies-list-fetch-movies/)


anastasiia-tilikina (changes requested):
this is how your demo page looks like right now =)

image

let's fix it and then we can move on

BipBobBoop:
> this is how your demo page looks like right now =)
> 
> image
> 
> let's fix it and then we can move on

On the left is an example from the assignment, on the right is mine. I didn't understand what is wrong?

image

VitaliyBondarenko1982 (approved):
Looks good. 
App works, after find movie it add to the list it appears on left side.

PR #1168, opened 2023-10-14
fetch movies

donizer:
[DEMO LINK](https://donizer.github.io/react_movies-list-fetch-movies/)


donizer:
GitHub fails to start test. But locally all passes.

PR #1179, opened 2023-10-27
done

maxonshev:
https://maxonshev.github.io/react_movies-list-fetch-movies/


maxonshev:
Everything works well and the tests pass, but there's one thing:

If we find a movie, for example, Inception. And after that, we press Search Again, without changing the title, the entire container block reloads.

But in the example, this doesn't happen.
How can I achieve the same effect?

PR #1586, opened 2025-03-24
Develop

AnOvramenko:
[DEMO LINK](https://AnOvramenko.github.io/react_movies-list-fetch-movies/)

### DIFF
--- a/src/components/FindMovie/FindMovie.tsx
+++ b/src/components/FindMovie/FindMovie.tsx
@@ -5,7 +5,7 @@ import { Movie } from '../../types/Movie';
 import { getMovie } from '../../api';
 import { MovieCard } from '../MovieCard';
 interface Props {
-  setNewMovie: (movie: (prev: Movie[]) => Movie[]) => void;
+  setNewMovie: (movie: Movie) => void;
 }
 const defaultImg = 'https://via.placeholder.com/360x270.png?text=no%20preview';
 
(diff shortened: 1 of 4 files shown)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,70 @@
...
+        setFindError(false);
+        setMovie(newMovie);
+      })
+      .finally(() => {
+        setIsLoading(false);
+      });
+  };
+
+  const handleAddMovie = (newMovie: Movie) => {
+    setNewMovie((prevListOfMovies: Movie[]) => {
```
daniil-trusov: `setNewMovie` body should be in `App` component. `FindMovie` is responsible for load single movie by query and send it as prop to `setNewMovie` callback.

daniil-trusov (changes requested):
Good job overall, however you have some logic in FindMovie that should be in the App component. Please check the comment for details and fix it.

daniil-trusov (approved):
Now it's all correct, good job!

src/components/FindMovie/FindMovie.tsx
AnOvramenko: setNewMovie it was a setState func from useState hook, so technically it works in app component))
But I agree that for better readability, this way looks much better

PR #1648, opened 2025-06-27
add solution

zbvika:
[DEMO LINK](https://zbvika.github.io/react_movies-list-fetch-movies/)


src/App.tsx
```diff
@@ -5,7 +5,19 @@ import { FindMovie } from './components/FindMovie';
...
+  function addToTheList(movie: Movie) {
+    setMovies(prev => {
+      const alreadyAdded = prev.some(mov => mov.imdbId === movie.imdbId);
+
+      if (alreadyAdded) {
+        return prev;
+      }
+
+      return [...prev, movie];
+    });
```
BudnikOleksii, line 19:
```suggestion
    const hasMovie = prev.some(mov => mov.imdbId === movie.imdbId);

    if (hasMovie) {
        return;
    }

    setMovies(prev =>[...prev, movie]);
```
You can do the check before setting a state

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,10 +1,53 @@
...
+    setTitle(event.target.value.trimStart());
+    clearError();
+  }
+
+  function handleSubmit(event: React.FormEvent) {
+    event.preventDefault();
+
+    const trimedTitle = title.trim();
+
+    if (trimedTitle === '') {
```
BudnikOleksii, line 33:
```suggestion
    if (!trimedTitle) {
```
You don't need to compare with an empty string, just check if a value exists

src/hooks/useMovies.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { getMovie } from '../api';
+import { MovieData } from '../types/MovieData';
+import { Movie } from '../types/Movie';
+
+export const useMovie = () => {
+  const [preMovie, setPreMovie] = useState<Movie | null>(null);
+  const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [isError, setIsError] = useState<boolean>(false);
+
+  function getNormalMovie(movieFromServer: MovieData): Movie {
```
BudnikOleksii, line 11:
```suggestion
  function normalizeMovie(movieFromServer: MovieData): Movie {
```
You can move this function to utils/helpers file/folder

BudnikOleksii (approved):
👍

PR #1661, opened 2025-08-05
add a solution

viktoriamyhailiak:
https://viktoriamyhailiak.github.io/react_movies-list-fetch-movies/


Anton-Kuchmasov (changes requested):
Great work!

To improve:
1. Input should NOT be in red error frame when App is initially loaded:
image
2. Small fetch logic improvement (check out the comment in FindMovie Component)

src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,7 +1,71 @@
...
+              imdbUrl: `https://www.imdb.com/title/${result.imdbID}`,
+              imdbId: result.imdbID,
+            };
+
+            setMovie(elem);
+          }
+        }
+      })
+      .finally(() => setIsLoading(false));
+  };
```
Anton-Kuchmasov, line 55: Actually, you don't need to check if ('Title' in result) — it seems unnecessary, since we've already checked if (Response === 'True').

Anton-Kuchmasov (approved):
AI approved your solution before - but please, check out the comments! :)

PR #1686, opened 2025-09-05
Added solution

YevheniiKa:
[DEMO LINK]
https://YevheniiKa.github.io/react_movies-list-fetch-movies/


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,124 @@
...
+    setMovie(null);
+    setErrorMessage('');
+
+    try {
+      const data = await getMovie(query);
+
+      if ('Response' in data && data.Response === 'False') {
+        showError("Can't find a movie with such a title");
+      } else {
+        setMovie(mapMovie(data as MovieData));
```
alexander-ignatow: You should fix getMovie - the result should be cast to the correct type inside
Or you need to check for the type ResponceError.

Anyways - your final result should not include excplicit casting

```suggestion
        setMovie(mapMovie(data));
```

alexander-ignatow (approved):
Good, but review my comment

And more - you probably should add your demo URL to some omdbapi list - as it fails with the CORS error

src/components/FindMovie/FindMovie.tsx
YevheniiKa: > You should fix getMovie - the result should be cast to the correct type inside Or you need to check for the type ResponceError.
> 
> Anyways - your final result should not include excplicit casting

Thanks for highlight, fixed in [02a5fb8](https://github.com/mate-academy/react_movies-list-fetch-movies/pull/1686/commits/02a5fb8a4c4fd5d0a277dcd85b34f7063ace4861)

PR #1776, opened 2026-04-10
Implement the FindMovie component to load movies from OMDb API

Dianaeva:
[DEMO LINK](https://dianaeva.github.io/react_movies-list-fetch-movies/)


src/components/FindMovie/FindMovie.tsx
```diff
@@ -1,57 +1,132 @@
...
+  };
+
+  const handleAddMovie = (movie: Movie) => {
+    addMovie(movie);
+    setTitle('');
+    setHasError(false);
+    setFoundMovie(null);
+  };
+
+  const movie = foundMovie && prepareMovie(foundMovie);
```
polosanya, line 64: Would be great to use useMemo to avoid recreating movie on each render

polosanya (approved):

